PR #39, opened 2019-12-13
fix(mainwindow): use simplified instead of trimmed in diff

### LINKED ISSUE #38: Use testlib.h to check the verdicts

Now the diff viewer works incorrectly for me, I guess it's about the newline character (I'm using Windows 10, and this happens only when there are more than one lines). And I have to follow a certain format strictly, even the space characters.

Use testlib.h to check the verdicts. Users can choose one of [the default checkers](https://github.com/MikeMirzayanov/testlib/tree/master/checkers) by a drop-down box, and [wcmp](https://github.com/MikeMirzayanov/testlib/blob/master/checkers/wcmp.cpp) is the default one.

The details besides AC/WA can be shown in the "Compiler Messages and stderr" section.

Use something like `diff -b` to ignore spaces.

**Additional context**

N/A

ouuan:
This partially fixes #38.

### DIFF
--- a/src/mainwindow.cc
+++ b/src/mainwindow.cc
@@ -823,7 +823,7 @@ void MainWindow::firstExecutionFinished(QString Stdout, QString Stderr) {
     Log::MessageLogger::error("Runner[1]:[STDERR]", Stderr.toStdString(), true);
   if (Stdout.isEmpty() || expected1->isEmpty())
     return;
-  bool isSame = Stdout.trimmed() == expected1->trimmed();
+  bool isSame = Stdout.simplified() == expected1->simplified();
 
   if (isSame)
     updateVerdict(Core::Verdict::ACCEPTED, 1);
@@ -839,7 +839,7 @@ void MainWindow::secondExecutionFinished(QString Stdout, QString Stderr) {
     Log::MessageLogger::error("Runner[2]:[STDERR]", Stderr.toStdString(), true);
   if (Stdout.isEmpty() || expected2->isEmpty())
     return;
-  bool isSame = Stdout.trimmed() == expected2->trimmed();
+  bool isSame = Stdout.simplified() == expected2->simplified();
 
   if (isSame)
     updateVerdict(Core::Verdict::ACCEPTED, 2);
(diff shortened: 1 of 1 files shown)


coder3101:
https://doc.qt.io/qt-5/qstring.html#simplified

According to docs this function replaces `\n` characters. So it basically makes everything one line, every multiline output is converted into one line output. So now following are same :

```
2
4 
6
```

and 

```
2 4 5

```

While this is obviously Wrong answer on Codeforces or any other platforms.

*Reverted the commit, Please open a new PR fixing this issue* (We need to manually implement that function I talked about in #38)

ouuan:
> answer

In other OJs this may be WA, but in most problems on CF, white spaces don't matter.

ouuan:
[image](https://docs.google.com/document/d/e/2PACX-1vRhazTXxSdj7JEIC7dp-nOWcUFiY8bXi9lLju-k6vVMKf4IiBmweJoOAMI-ZEZxatXF08I9wMOQpMqC/pub)

coder3101:
Yes, So people who use Codechef and other platforms will complain about it. It will be better if newlines are maintained.

ouuan:
Where should I write a function for comparing? It doesn't seem like a "member function".

coder3101:
Create a function with following signature in mainwindow.cc.
```cpp
bool MainWindow::isVerdictPass(QString output, QString expected){...};
```

It should return true if verdict is `AC` otherwise false.
Create this after the definition of function `launchSession()`. 

Then you need to declare it as well, in mainwindow.hpp, just after launchSession(); declare it (private);

```cpp
....
bool isVerdictPass (QString, QString);
...
```

Then replace that `isSame` with the value returned from this call, pass the arguments as expected.

coder3101:
> Where should I write a function for comparing? It doesn't seem like a "member function".
> 
> (In fact, almost all functions returns void.)

Yes they are member function that behave like callbacks, callbacks usually have void return type. They are normal C++ classes and functions.  No magic, you can do whatever you like.

Issue #23, opened 2019-12-06
Save when compiling / closing warning only when the file is not saved


ouuan:
**Is your feature request related to a problem? Please describe.**

I usually forget to save the file, because I'm used to other IDEs which automatically save files when compiling, and warn you at closing only when the file is not saved.

**Describe the solution you'd like**

I think the file should be saved when compiling, or make it an option. Maybe only be enabled when the corresponding file is specified, which allows compiling without the file like the online IDEs.

Also, I think the warning when closing the editor should only be there when the file is not saved.

Something represents for "the file has been modified" is also helpful, like:

![image](https://user-images.githubusercontent.com/30581822/70326387-6c8f7980-186f-11ea-9727-0f1ae0a9accd.png)

![image](https://user-images.githubusercontent.com/30581822/70326372-66010200-186f-11ea-96b2-e32c37653c90.png)

**Describe alternatives you've considered**

Different warnings for normal exiting / exiting without saving.

**Additional context**

N/A

coder3101:
When you start a new session the app automatically saves the current buffer.

coder3101:
Fixed. Now Editor saves the buffer each time compilation is requested.

ouuan:
> Also, I think the warning when closing the editor should only be there when the file is not saved.
>
> Something represents for "the file has been modified" is also helpful, like:
>
> ![image](https://user-images.githubusercontent.com/30581822/70326387-6c8f7980-186f-11ea-9727-0f1ae0a9accd.png)
>
> ![image](https://user-images.githubusercontent.com/30581822/70326372-66010200-186f-11ea-96b2-e32c37653c90.png)

This part is not fixed yet.

PR #44, opened 2019-12-13
feat(mainwindow): update saving logic & detect text change

ouuan:
This fixes the second part of #23.

It's a big change, I'm not so sure that there aren't any mistakes. Please check it carefully.

BTW, the code reuse is terrible. There are too many copy pastings, especially the 3 inputs, 3 outputs, 3 runners part.

### DIFF
--- a/include/mainwindow.hpp
+++ b/include/mainwindow.hpp
@@ -44,6 +44,8 @@ class MainWindow : public QMainWindow {
   ~MainWindow() override;
   void closeEvent(QCloseEvent* event) override;
  private slots:
+  void on_textChanged_triggered();
+
   void on_actionQuit_triggered();
 
   void on_actionDark_Theme_triggered(bool checked);
@@ -134,10 +136,14 @@ class MainWindow : public QMainWindow {
   void restoreSettings();
   void setupCore();
   void launchSession();
-  bool isVerdictPass(QString, QString);
   void launchCompanionSession(Network::CompanionData);
   void checkUpdates();
   void createAndAttachServer();
   void updateVerdict(Core::Verdict, int);
+
+  bool isVerdictPass(QString, QString);
+  bool saveFile(bool);
+  bool isTextChanged();
+  bool confirmUnchanged();
 };
 #endif  // MAINWINDOW_HPP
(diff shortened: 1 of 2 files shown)


coder3101:
Yes lot of code was inherited from my old editor. Which was for personal and fun use! I will tr making it more organised !!

ouuan:
Something should be changed in favor of `launchCompanionSession`.

coder3101:
> Something should be changed in favor of `launchCompanionSession`.

What?

ouuan:
> > Something should be changed in favor of `launchCompanionSession`.
> 
> What?

I would fix it. Left a comment here for:
1. self reminding;
2. don't merge.

coder3101:
When your work is done, mention me to merge this work!

coder3101:
Application terminates because it thinks that no window is visible. 

Read this problem and in the problem itself is the answer. 

Basically what you have to do is call `app.setQuitOnLastWindowClosed(false)` in main as soon as app is created. 

Read this answer https://stackoverflow.com/questions/5116459/problem-with-hidden-qmainwindow-application-crashes-after-qmessagebox-is-displa

ouuan:
> Application terminates because it thinks that no window is visible.
> 
> Read this problem and in the problem itself is the answer.
> 
> Basically what you have to do is call `app.setQuitOnLastWindowClosed(false)` in main as soon as app is created.
> 
> Read this answer https://stackoverflow.com/questions/5116459/problem-with-hidden-qmainwindow-application-crashes-after-qmessagebox-is-displa

The bug still happens:

```cpp
int main(int argc, char* argv[]) {
  QApplication a(argc, argv);
  a.setQuitOnLastWindowClosed(false);
  QStringList args = a.arguments();
  QString filePath = args.size() > 1 ? args[1]: "";
  MainWindow w(filePath);
  w.show();
  return a.exec();
}
```

ouuan:
How about this: open a new session when it's an unchanged temporary buffer (i.e. unsaved, the content is exactly the template), otherwise don't open a new session.

ouuan:
@coder3101 

It looks good to me now, you can merge it after checking again.

coder3101:
After i merge it today, i will release a pre-release version, `can you guarantee that New session won't crash application?`

ouuan:
> After i merge it today, i will release a pre-release version, `can you guarantee that New session won't crash application?`

It didn't crash when I tested it, and in this version, there is no message box opened in the backstage.

PR #48, opened 2019-12-15
fix(mainwindow): update save as

ouuan:
1. fix the bug of saving a file as itself.
2. Use saveFile() for save as, more code reuse.
3. change openFile to the file saved as, which is the same as many other editors/IDEs.


coder3101:
I am adding CI so that AppImage issue with Linux could be solved. Meanwhile, False Build fails are expected!

PR #53, opened 2019-12-16
fix(MessageLogger): use append instead of insertHtml 

ouuan:
In the old version, if you click in the messageLogger, the place of the "cursor" will change, and new logs will be inserted after the "cursor".


coder3101:
How exactly does this changes message logger?

ouuan:
It fixes the bug I described in the commit message.

ouuan:
You can open auto-save and click on the text of the log in the old version to reproduce the bug.

PR #54, opened 2019-12-16
chore: add download link for Chinese users

ouuan:
Downloading on Github can be extremely slow (below 10KiB/s) in Mainland China, this link may be useful for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Head over to the [release page](https://github.com/coder3101/cp-editor2/releases
   >See google on how to run AppImages if you are new to AppImages.
 - Mac users please install from source for the time being.
 
-For Chinese users who are not convenient to download on Github, there is an [alternative download link](https://pan.wzf2000.top/s/md70l8h0).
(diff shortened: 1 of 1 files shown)


coder3101:
So is it a mirror?

ouuan:
No, it's uploaded by myself.

coder3101:
So, every time a new release comes! This url will change? You must gaurantee that the binary in the above url will be latest always!

ouuan:
The URL won't change (it will be still the same after changing the content), but manually upload is required.

coder3101:
Okay, then you will have to upload the binaries from releases page as soon as a stable release hits. Should I merge?

ouuan:
I'll try to upload it ASAP, or you can register on that website and upload by yourself. What do you think?

coder3101:
I am setting Automation tools to do that for me. Anyways! Here is what you can do, keep that link and let user know that the exe downloaded from that place may not be updated. It will be easy for someone, to quickly try it after downloading then if they like they can get the latest version by visiting the release page.

PR #58, opened 2019-12-24
fix(mainwindow.cc): restore save tests setting

ouuan:
I think it should be my mistake when using git. I remember this setting is restored when I tested it locally. 😢


ouuan:
Oh, I was stupid. There was nothing wrong with my use of git. The setting is saved, only the check wasn't restored.

coder3101:
I have already released it on 3.2.1

PR #67, opened 2020-01-02
format codes & fix format shortcut & update QCodeEditor

ouuan:


coder3101:
I assume this code format is done by `.clangformat` in source root?

ouuan:
> I assume this code format is done by `.clangformat` in source root?

I think so. Visual Studio said it used `.clangformat` in source root automatically.

ouuan:
Maybe the QCodeEditor should be separated? I added it together with formatting carelessly.

ouuan:
But I have to go to the dormitory now :(

coder3101:
You don't need to change QCodeEditor files in PR. Instead you should update it locally with git submodule update command. I will not merge it now, you need to remove that QCodeEditor file from this PR rest is Okay. Take your time, soon i will publish the new blog about this editor. Meanwhile you could try it and so this is why it's in beta.

ouuan:
BTW, what's the problem with the system-wide dark theme? (I don't know where to ask this question, so ask here.)

coder3101:
> BTW, what's the problem with the system-wide dark theme? (I don't know where to ask this question, so ask here.)

Qt does not support Dark theme, the current theme is nothing but a Qt implementation of KDE that has dark theme. In KDE Plasma and other distro that mix well with Qt, you can turn on system theme Dark and it manages and changes the theme to dark. I will however look into if there is a way out.

coder3101:
Hey, if you get some time could you also re-write(modify) the README to match the new Editor features?

ouuan:
> > BTW, what's the problem with the system-wide dark theme? (I don't know where to ask this question, so ask here.)
> 
> Qt does not support Dark theme, the current theme is nothing but a Qt implementation of KDE that has dark theme. In KDE Plasma and other distro that mix well with Qt, you can turn on system theme Dark and it manages and changes the theme to dark. I will however look into if there is a way out.

I mean, what are the "serious UI/UX issues", and why was it available in old versions but not in the current version.

PR #70, opened 2020-01-08
feat(MessageLogger): make messages HTMLescaped

ouuan:
Compiler messages like `vector<int>` should be HTMLescaped.


ouuan:
> So what does this do better? For safety or does it really makes something good?

See the commit message. Compiler messages like `vector<int>` was like `vector` before.

PR #71, opened 2020-01-08
Some changes about tabs

ouuan:
See the commit messages for more information.

### DIFF
--- a/include/SettingsManager.hpp
+++ b/include/SettingsManager.hpp
@@ -29,7 +29,8 @@
 namespace Settings
 {
 
-enum ViewMode{
+enum ViewMode
+{
     FULL_EDITOR,
     FULL_IO,
     SPLIT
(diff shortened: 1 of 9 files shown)


ouuan:
I'm solving a segment fault introduced in 358c3fe, please wait before merge.

ouuan:
To reproduce the segmentation fault:

1. open the editor
2. open a new tab (then there are two tabs, and the current tab is the second one)
3. close the first tab

If you switch to the first tab and close the second one, everything works well.

Deleting https://github.com/coder3101/cp-editor/blob/358c3feaacff46b1807f0697dd7526b1fb9a245b/src/appwindow.cpp#L230 will solve this segmentation fault (but I can't delete it). But calling `tmp->disconnect()` before closing it won't solve this segmentation fault.

Do you have ideas on how to solve this?

ouuan:
I think I got the reason, switch to a tab will cause a `editorTextChanged`, and the `windowIndex` has not been updated yet.

ouuan:
OK, you can merge it.

PR #74, opened 2020-01-09
feat(appwindow.cpp): use default language

ouuan:


coder3101:
I have already chnged it myself and some other changes because your work was not able to open multiple tabs if companion parsed Complete contests. Also if same contest is parsed a duplicate tab won't be created anymore. Please check it yourself after compiling my current changes.

PR #75, opened 2020-01-09
Add snippets

ouuan:
I think the editing part is done, maybe needs some testing.

I'll add the "use snippets in the code editor" part tomorrow. Though it should be easy to implement, I don't have time today.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ build/
 
 # Visual Studio
 .vs
+
+# Visual Studio Code
+.vscode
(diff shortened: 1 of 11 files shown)


coder3101:
My work on CF tools to submit is almost done and no more big code changes will be required. Here is what i would like to also let you know, add snippets for all 3 language and it will be better if they are all 3 in different files. Maybe json files for each language snippets.

coder3101:
I just reviewed your code and it looks good. You can ignore that json to store snippets, .ini will be good to. All Editor related things in one place, i also see you implementing different languages snippets separately depending upto tab language. That is good as expected. Take your time and perfect this feature. Also Thanks for your valuable work in CP editor.

coder3101:
So i am merging this to master now. Also my work on cf-tools is ready as well. You could try now.

PR #87, opened 2020-01-11
refactor: use QSettings::clear() for reset settings

ouuan:
Sinces all defaults are in the "get" functions, a simple clear()
should do the resetting well.

What's more, this fixes the bug that unused settings won't be cleared.

For example, there is still "compile" without "cpp" in my ini file.


coder3101:
Please resolve this conflict.

Issue #92, opened 2020-01-11
Support of Hot Exit


m3h3d1:
**Is your feature request related to a problem? Please describe.**
CP-Editor doesn't remember the previous session.

**Describe the solution you'd like**
Enable the editor to exit immediately without prompting, preserving changes and unsaved files for the next time it is run. When the CP-Editor is launched again, all opened and edited files will be at the same state as they were before close. 
N.B: VS Code, Sublime Text supports hot exit.

**Describe alternatives you've considered**
If it is hard to code "Hot Exit" then firstly add support of remembering the previous tabs opened earlier. 

**Additional context**
N/A

coder3101:
@ouuan  need your comments here. It is easy to implement maybe you could guide someone here.

ouuan:
Should I guide him to make a PR?

ouuan:
I'd like to implement it myself though.

Issue #82, opened 2020-01-10
Put the compiled executable in the same folder as the source file


ouuan:
**Is your feature request related to a problem? Please describe.**

> Any decent antivirus scans just-made executables, so every time I compile it scans the .exe, which means I have to wait 10 seconds before it can actually run. Every single time I compile. The thing is, you can add an exception in the antivirus, but I'd prefer not adding an exception in my damn AppData folder!

**Describe the solution you'd like**

> Put the .exe in the same folder as the source file.

**Describe alternatives you've considered**

> At least let us choose where it should be.

**Additional context**

Originally aked [here](http://codeforces.com/blog/entry/71907?#comment-562850).

ouuan:
Also, I suggest using [QTemporaryFile](https://doc.qt.io/qt-5/qtemporaryfile.html) instead of `Base::Files`. `Base::Files` and `MainWindow::windowIndex` can be got rid of completely. Not only because it's more convenient and safe, the current solution is bad also because of the window index changes after closing/moving tabs.

I'd like to implement this after #96 and #97.

ouuan:
For self reminding: use `waitForFinished()` instead of `waitForStarted()` to get rid of unnecessary `kill()` and the warning message `QProcess: Destroyed while process ("g++") is still running.`.

ouuan:
I'd like to implement #92 first now. Reimplement compiler/runner/cf-tool is a big work, since I need to keep the temporary file in order to reuse the executable files, and I need to use different logic for untitled tabs, saved files, unsaved files with a filePath, save error, etc.

ouuan:
But I'll be thinking of "how to restore the window index" when I work on hot exit...

Seems that this issue is really a dependency of #92.

ouuan:
What about only reuse executable files if the source file is saved? That will make things easier because I can delete all temporary files right after using them.

coder3101:
Yes that would be nice tok

PR #97, opened 2020-01-14
feat: make file logic like Sublime

ouuan:
1. Open file when necessary instead of holding it.

2. Watch file changes on disk.

3. Show complete path when multiple tabs are of the same name.

4. Change language when saving unsaved files / saving as.

5. Text changes and other minor changes.

Please test these changes. I think there might be some bugs, especially when handling read/write file fails, and things related to problemURL.

### DIFF
--- a/include/appwindow.hpp
+++ b/include/appwindow.hpp
@@ -55,7 +55,7 @@ class AppWindow : public QMainWindow
 
     void onTabCloseRequested(int);
     void onTabChanged(int);
-    void onEditorTextChanged(bool, MainWindow *);
+    void onEditorChanged(MainWindow *);
     void onSaveTimerElapsed();
     void onSettingsApplied();
     void onSplitterMoved(int, int);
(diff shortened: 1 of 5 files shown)


ouuan:
I'd like to reimplement runner/compiler/cf-tool related with #82 first, then implement #92, for fewer conflicts.

coder3101:
Yes you are right, when i started the project i was not very aware of Qt API like QTemporaryFile. But Since now you have discovered it, it surely is nice to get rid of Base::Files because they can in future become a bottleneck and are not very portable. Let's Qt handle all cross platform things.

coder3101:
I guess this is work in progress. So I will review it once you ask me to do so.

ouuan:
> I guess this is work in progress. So I will review it once you ask me to do so.

I can also make a new PR and merge this to it, or work on #82 in this PR. But for a clean commit history, I'd like to have this PR merged first.

PR #99, opened 2020-01-15
Revert "feat: make file logic like Sublime"

coder3101:
Reverts coder3101/cp-editor#97


ouuan:
Do you want to make a release without this PR first and add it back later?

coder3101:
Yes exactly

PR #122, opened 2020-01-25
Attempt to fix cf-tool can't find

testitem:
I have literally zero knowledge of qt so I can't make a fancy dialog to ask the user what the cf-tool path is, but I think this *should* work on my system and most other systems. It's generally advised to put user binaries in /usr/local/bin and not /usr/bin anyways.

### DIFF
--- a/src/cftools.cpp
+++ b/src/cftools.cpp
@@ -63,7 +63,7 @@ void CFTools::submit(const QString &filePath, const QString &url, const QString 
         }
     }
 
-    CFToolProcess->setProgram("cf");
+    CFToolProcess->setProgram("/usr/local/bin/cf");
     CFToolProcess->setArguments({"submit", problemContestId, problemCode, filePath});
     connect(CFToolProcess, SIGNAL(readyReadStandardOutput()), this, SLOT(onReadReady()));
     CFToolProcess->start();
@@ -72,7 +72,7 @@ void CFTools::submit(const QString &filePath, const QString &url, const QString 
 bool CFTools::check()
 {
     QProcess checkProcess;
-    checkProcess.start("cf --version");
+    checkProcess.start("/usr/local/bin/cf --version");
     bool finished = checkProcess.waitForFinished(2000);
     return finished && checkProcess.exitCode() == 0;
 }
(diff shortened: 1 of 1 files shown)


ouuan:
This doesn't make sense since you can't force users to put the executable in `/usr/local/bin`, what's more it's not possible on Windows.

coder3101:
With just one code changes, i will merge this PR, See how Open File opens a dialog and it returns a path. Allow user to choose CF path , its just 5-10 lines of code nothing fancy. Then we will help you how to make this changes to preferences dialog or possibly. @ouuan could do settings part and let you know.

ouuan:
Why do you think all Mac users should put it in `/usr/local/bin`?

And you'd better not use anything from the std namespace, for `std::string`, use [QString](http://doc.qt.io/qt-5/qstring.html) instead.

ouuan:
@coder3101 It's weird that the bot removed `change_requested` tag right after a new commit without reviewer's approve.

coder3101:
I will re-request it. Please bear with bot, it was made by me long time ago when I was still learning Java script.

testitem:
@ouuan because that's the only place in path you can put it into (you can't put it into `/usr/bin` since everything in `/usr` outside of `/usr/local` is locked, even to root)

testitem:
wait what you made the bot

ouuan:
> @ouuan because that's the only place in path you can put it into (you can't put it into `/usr/bin` since everything in `/usr` outside of `/usr/local` is locked, even to root)

OK then, I've never used Mac...

ouuan:
@testitem To add a setting, you need to modify `ui/preferencewindow.ui`, `src/preferencewindow.cpp`, `include/SettingsManager.hpp`, `src/SettingsManager.cpp`, and apply the setting in `src/cftools.cpp`.

See cdabd44 for reference.

testitem:
I'll get to it tomorrow, it's getting a bit late. But I think the branch is in a mergeable state.

coder3101:
Resolved with 1a49958

It is recommended to please see how it is done to help and contribute in the project.

PR #123, opened 2020-01-25
Merge Development branch for next release 6.0

coder3101:
This branch will release CP Editor 6.0. This PR will always check merge-ability of this branch against master. As of now, `dev` is very upto-date with master. All critical changes of master have been merged in `dev`.

**Any immediate change to `master`, if brought directly should get merged to `dev` as well.**


ouuan:
Seems that everyone has its own style of the version number. I'd like this to be 5.4.0 and the current 5.4.0 to be 5.3.3, and release 6.0 after QScintilla is done. However, make this 6.0 is not bad. Some applications has the version number `0.4x.x`, we just move the zero from the front to the back.

coder3101:
I will add scintilla to dev branch first and only then release 6.0

ouuan:
> I will add scintilla to dev branch first and only then release 6.0

How long will it take? I think whatever the version number is, the next beta release (includes the current master only, if scintilla needs some time) should be out ASAP.

PR #125, opened 2020-01-26
Test Azure

coder3101:
This PR builds Azure on MSVC. When merged the setup will be created by Azure and released by its own.


ouuan:
For reference, I use this on WIndows:

```
cmake ..
cmake --build . --config Release
```

coder3101:
Its hard to install qt on windows via CMD then add it to path and deploy. I will create some dummy released, to test this.

ouuan:
Try to output some messages like `echo %CMAKE_PREFIX_PATH%`, `dir $(Build.BinariesDirectory)`, `dir $(Build.BinariesDirectory)\Qt\5.14.0\msvc2017\lib\cmake`?

coder3101:
That environment variable is being set. Okay, I will try it now. Also don't worry this all commits will be squashed into one before merge to master.

Issue #151, opened 2020-01-31
Avoid unnecessary copies, pass by const ref


coder3101:
**Is your feature request related to a problem? Please describe.**
At many places the functions take things as copy and this adds to performance. In functions where it is possible to have pass by const references should take argument by const ref. Most of such cases are reported by clang-tidy.

**Describe the solution you'd like**
Refactor functions to take argument that could be passed by const ref.

**Describe alternatives you've considered**
N/A

**Additional context**
One copy of say SettingsData class might seem insignificant but if it's called 1000 times during program execution. The whole things surely adds some performance overhead. Using clang-tidy we can enhance our code quality and increase performance as well.

testitem:
Hey, I can try to do this

testitem:
wait what why did it unassign ouuan

coder3101:
Sure give it a try. We will suggest changes in PR section of any required. This is also a good way to learn how editor works.

testitem:
What is wrong with the assign system!

testitem:
I will make all the functions in SettingsManager.cpp inlined, as just calling the function incurs a significant overhead if not compiled with -O3.

Furthermore, for one-liner functions, I prefer

```
bool SettingManager::isMaximizedWindow() { return mSettings->value("win_max", "false").toBool(); }
```

instead of

```
bool SettingManager::isMaximizedWindow()
{
    return mSettings->value("win_max", "false").toBool();
}
```

, is that OK?

coder3101:
You can make them inline but don't change formatting. It is there for consistency and better readability

coder3101:
Also You may need to re-format code with clang-format. Using the .clang-format file in root of project

ouuan:
> I will make all the functions in SettingsManager.cpp inlined, as just calling the function incurs a significant overhead if not compiled with -O3.
> 
> Furthermore, for one-liner functions, I prefer
> 
> ```
> bool SettingManager::isMaximizedWindow() { return mSettings->value("win_max", "false").toBool(); }
> ```
> 
> instead of
> 
> ```
> bool SettingManager::isMaximizedWindow()
> {
>     return mSettings->value("win_max", "false").toBool();
> }
> ```
> 
> , is that OK?

I don't think it should be inlined because it's seldom used.

coder3101:
> I don't think it should be inlined because it's seldom used.

[...]

PR #154, opened 2020-02-01
refactor: const ref most functions in SettingsManager & inline, and fix compiler warnings

testitem:
## Description

 1. All functions in SettingsManager.cpp that were one line (or were made one line, with a new function Util.hpp/bool2string) were inlined into SettingsManager.hpp.
 2. All corresponding parameters that can be made a const ref were made a const ref.
 3. Fixed multiple compiler warnings.
 4. Fixed typo in copyright header in all files.

## Related Issue

#151 

## Motivation and Context

Performance

## How Has This Been Tested?

On MacOS Mojave with Brew QT 5.12

## Type of changes

- [ ] Bug fix (a change which fixes an issue)
- [ ] New feature (a change which adds functionality)
- [ ] Documentation (a change which modifies the documentation only)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/coder3101/cp-editor/blob/master/CONTRIBUTING.md) document.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,9 +20,14 @@
 
 ## Type of changes
 <!--- What type of changes does your code introduce? Put an `x` in the box that applies: -->
-- [ ] Bug fix (a change which fixes an issue)
-- [ ] New feature (a change which adds functionality)
-- [ ] Documentation (a change which modifies the documentation only)
+- [ ] Bug fix (changes which fix an issue)
+- [ ] New feature (changes which add functionality)
+- [ ] Documentation (changes which modify the documentation only)
+- [ ] Style change (changes which do not affect the meaning of the code: code formatting, etc.)
+- [ ] Refactor (changes which affect the meaning of the code but neither fix a bug nor add a feature)
+- [ ] Performance improve (changes which improve performance)
+- [ ] Test (changes which add tests)
+- [ ] Chore (changes which do not belong to any type above)
 
 ## Checklist
 <!--- Go over all the following points, and put an `x` in all the boxes that apply. -->


include/SettingsManager.hpp
```diff
@@ -95,145 +97,182 @@ struct SettingsData
...
 class SettingManager
 {
+    // All functions with code here are implicitly inlined by the compiler.
   public:
     SettingManager();
 
-    int getConnectionPort();
-    void setConnectionPort(int port);
+    int getConnectionPort() { return mSettings->value("companion_port", 10045).toInt(); }
+    void setConnectionPort(const int &port) { mSettings->setValue("companion_port", port); }
```
ouuan: Don't use `const &` for basic types. In fact, compilers might optimize it to pass by value.

include/Util.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef UTIL_HPP
+#define UTIL_HPP
+
+#include <QString>
+
+inline QString bool2string(const bool &value)
+{
+    if (value)
+        return QString::fromStdString("true");
```
ouuan:
```suggestion
        return QString("true");
```

include/Util.hpp
```diff
@@ -0,0 +1,31 @@
...
+    else
+        return QString::fromStdString("false");
```
ouuan:
```suggestion
        return QString("false");
```

include/SettingsManager.hpp
ouuan: It's not necessary, but without optimization, this might be slower than pass by value, because the value is smaller than a pointer.

testitem: Well, in fact it's going to be inlined, so const & won't really make a difference here. The function is implicitly inlined.

ouuan:
@testitem You can read [this](https://chris.beams.io/posts/git-commit/) for how to write commit messages.

testitem:
What does "LGTM" mean?

testitem:
if(x) is more easier to understand and read than if(x != nullptr), since if x is nullptr it implicitly evaluates to false and it's nicer to read as "if x exists".

testitem:
I, as a former Python user, firmly believe that if code can be more concise without hurting understandableness, then the code should be more concise. "If(something != 0)" is just redundancy.

ouuan:
You can remove `== nullptr`, but please keep `== 0`.

See also: https://github.com/isocpp/CppCoreGuidelines/issues/1006#issuecomment-317848446

ouuan:
However, I still think `x != nullptr` also suggests `x` is a pointer.

ouuan:
I'm not sure about it, maybe we need more discussions and references.

testitem:
Right now, we have two options:

 1. Stick to a as-concise-as-possible code, but needs a second of thinking (or context about the type of a variable)
 2. Stick with verbose code, making the code longer and potentially more difficult to read, but the types and usages of variables/functions are obvious

Poll? React +1 for 1, and -1 for 2.

ouuan:
I don't think 2 is more difficult to read, and as you say the type is clear.

testitem:
I think adding "!= nullptr" disrupts my line of thinking when reading code, and can be removed to make reading the code easier and fluent. But when "!= nullptr", it is more apparent that the compared object is a pointer, especially in "log != nullptr" when "log" is declared in a wholly different file.

ouuan:
image

testitem:
Fine, but if so I want to declare a macro "exists" that expands into "!= nullptr", so variables read as "if(log exists)".

quank123wip:
I think when people reading a code people should know what is this meant, and how to make it work, instead of enjoying its beauty, so I think it would be better if we keep "!=nullptr", as C++ is a Strong type language, people would know x is a pointer or x is a variable as they see "x!=nullptr", that people don't need to find its declaration. :)

testitem:
Yeah, that's my point.

Are you the person in the QQ chat?

ouuan:
So please keep all `== 0` and `== nullptr`.

ouuan:
I pushed wrong commit and seems that it closes this PR and I can't reopen it now, please reset to e255d1d and force-push.

ouuan:
@testitem I made a mistake and closed this accidentally. Please push to e255d1d and push so that I can reopen this.

```sh
git reset --hard e255d1d
git push -f
```

coder3101:
Where did all commits of this PR go?

ouuan:
> Where did all commits of this PR go?

I mistyped `git push -f testitem dev:dev` instead of `git push -f testitem/dev:dev`. The commits are on the Github, in my local disk, and in his local disk. But I don't know where is he.

coder3101:
> I don't know where is he.

@testitem Are you there?

testitem:
no since i'm back to using acmx :|

ouuan:
> no since i found out how to use acmx :|
> 
> I will get to cp-editor when i have time, which is certainly not now because I have priorities in CP

So do you want to do a single push or let us open a PR? We'd probably squash and merge for a clear commit history, so if you don't push by yourself, you may lose the contribution.

PR #171, opened 2020-02-17
Test

ouuan:
Auto clang format test.


coder3101:
Make sure that clang-format does not changes "third_party" codes. Also let me know if you need Token with push access to repository.

ouuan:
I think a65ecb2 is OK, a personal token is not required, and auto format will be pushed to the branch of the formatted commit, pull requests will not affect the base branch.

coder3101:
Can we delete auto-clang-format branch?

PR #216, opened 2020-04-02
(WIP) Add C++ debugging support

### LINKED ISSUE #17: Add GDB support

I don't know how to use GDB with CP Editor.

Support GDB with GUI, like many other IDEs.

Support GDB in command line.

**Additional context**

N/A

neko-para:
## Description

Use gdb, gdbserver in path to debug c++ program. Add a simple dialog for debug control.
Add a new library [qgdbint](https://github.com/neko-para/qgdbint).
Still work in progress.

## Related Issue

This closes #17.

## How Has This Been Tested?

write simple program and check whether debug feature works well(including Excution, Breakpoints.)

* Ubuntu 18.04 x64(on WSL), gdb 8.1
* Windows/MSYS2 x64, gdb 9.0

## Type of changes

## Checklist

- [ ] I have used clang-format-9 and `.clang-format` file in the root directory to format my codes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "third_party/testlib"]
 	path = third_party/testlib
 	url = https://github.com/MikeMirzayanov/testlib
+[submodule "third_party/qgdbint"]
+	path = third_party/qgdbint
+	url = https://github.com/neko-para/qgdbint
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,8 @@ add_subdirectory(third_party/QtFindReplaceDialog/dialogs)
 
 add_subdirectory(third_party/lsp-cpp)
 
+add_subdirectory(third_party/qgdbint)
+
 configure_file(version.hpp.in ${CMAKE_BINARY_DIR}/generated/version.hpp)
 
 IF(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
@@ -88,6 +90,9 @@ add_executable(cpeditor
     src/preferencewindow.cpp
     src/appwindow.cpp
 
+    include/debugcontrol.hpp
+    src/debugcontrol.cpp
+
     include/Util.hpp
     src/Util.cpp
 
@@ -99,6 +104,7 @@ add_executable(cpeditor
     ui/mainwindow.ui
     ui/preferencewindow.ui
     ui/appwindow.ui
+    ui/debugcontrol.ui
 
     ${CMAKE_BINARY_DIR}/generated/version.hpp
 
@@ -109,6 +115,7 @@ include_directories("third_party/QCodeEditor/include")
 include_directories("third_party/diff_match_patch")
 include_directories("third_party/QtFindReplaceDialog")
 include_directories("third_party/lsp-cpp/include")
+include_directories("third_party/qgdbint")
 include_directories("include/")
 include_directories("generated/")
 
@@ -118,6 +125,7 @@ target_link_libraries(cpeditor PRIVATE QCodeEditor)
 target_link_libraries(cpeditor PRIVATE SingleApplication)
 target_link_libraries(cpeditor PRIVATE QtFindReplaceDialog)
 target_link_libraries(cpeditor PRIVATE LSPClient)
+target_link_libraries(cpeditor PRIVATE qgdbint)
  
 if(APPLE)
     set_target_properties(cpeditor
--- a/include/Core/Runner.hpp
+++ b/include/Core/Runner.hpp
@@ -130,6 +130,12 @@ class Runner : public QObject
     void onTimeout();
 
   private:
+    struct Command
+    {
+        QString prog;
+        QStringList args;
+        QString full;
+    };
     /*
      * @brief get the command to run a program
      * @param filePath the path to the source file
@@ -137,7 +143,7 @@ class Runner : public QObject
      * @param runCommand the command for running a program
      * @param args the command line arguments added at the back to start the program
      */
-    QString getCommand(const QString &filePath, const QString &lang, const QString &runCommand, const QString &args);
+    Command getCommand(const QString &filePath, const QString &lang, const QString &runCommand, const QString &args);
 
     const int runnerIndex;             // the index of the testcase
     QProcess *runProcess = nullptr;    // the process to run the program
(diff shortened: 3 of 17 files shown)


ouuan:
Did you use new features in Qt 5.14?

neko-para:
I use Qt::SkipEmptyParts, which was introduced in Qt 5.14. I'll fix this in the next commit.

ouuan:
BTW, you can [enable GitHub Actions in your fork](https://github.com/neko-para/cpeditor/actions), then your codes will be automatically formatted. When you are finished, I can squash and merge, and the format commits will be squashed into a single commit, which is better than merge into master and then another format commit.

neko-para:
I'll do it later.
Is it possible to change the old settings' type? I'd like to change the type of runtime arguments from string to stringlist, but this may cause some compability issue. Or it's ok to try to parse the string to strnglist, which only influence the debug(I use stringlist as arguments when excuting gdb)

neko-para:
I've just found that it's easy to install clang format 9 on ubuntu(I used to think that only old versions are in the repository). I'll do the formatting locally.

ouuan:
Are there any problems with the current implementation? I guess it's OK to use the runtime arguments (a QString) as an element of a  QStringList, or use QString in `QGdbProcessManager::prepare`.

neko-para:
This will combine all arguments into one, which is not what it should be.
For example, we supposed to run
./test arg1 arg2
In the current implementation, it works like
./test "arg1 arg2"
It may cause unknown problems(this won't happen in normal run. It only appear when debugging)

ouuan:
What about using QString in `QGdbProcessManager::prepare`?

neko-para:
It has the same problem.
Maybe i can use ```void QProcess::start(const QString &command, QIODevice::OpenMode mode = ReadWrite)``` instead of ```void QProcess::setArguments(const QStringList &arguments)```.
I'll try it later.

neko-para:
I've do some tests.
I use the start(cmd), and it doesn't work as expected.
when I run
./test arg1 "arg 2"
which means set arguments to ```arg1 "arg 2"``` and ./test will print argc
'run' will print 3 and 'debug' will print 4(which means the second argument is divided into two part)
But when i set argument to ```arg1 "arg\ 2"```, 'debug' print 3.
Maybe I should manually divide the argument string

coder3101:
> Or it's ok to try to parse the string to strnglist.

This is a better approach, Neither our QString will be affected nor Any settings changes are required. Make sure to chop them correctly handling "(quotes) and other special characters. 

Like,

```cpp
$ ./capitalizer "capitalize this" --upto 5
$ CAPITalize this
```

It has only 3 arguments. In settings, we will make sure that user adds a `" "` or `' '` for space-separated commands. This also makes sense, doesn't it? You would split it into QStringList as `{capitalize this, --upto, 5}` and use for your purpose.

BTW, Your code looks Okay to me, Just format it (optionally), and Make sure that You keep our source compilable on at least 5.12

Thanks for your contribution.

neko-para:
I write a function in Util and it works well. I'll also change the Runner to use the splitted arguments.

neko-para:
But i'm not sure the run_xxx settings should always be a program or could contain some option. If it contains options, I shall parse it too.

ouuan:
> But i'm not sure the run_xxx settings should always be a program or could contain some option. If it contains options, I shall parse it too.

I think options should be in the runtime arguments :thinking:

neko-para:
So the run command should always be a path to program? If the run command contains options like ```java -xxx```(I know few about java), I can parse it.

ouuan:
Is there any case that the options must be in the run command not in the runtime arguments?

neko-para:
I don't know. Actually when someone needs this feature, he should change the IDE

ouuan:
> I don't know. Actually when someone needs this feature, he should change the IDE

In fact, I don't know how runtime arguments are useful in Competitive Programming.

neko-para:
```
$ python --help
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
...
```
according the usage, I think there are many options should be placed before the file

neko-para:
yep. It is not so important

ouuan:
However, perhaps it's still good to make it able for users to add options both at the beginning (run command) and at the end (runtime arguments).

neko-para:
But this feature may cause compability problem.
In the old(and now) setting file, the program path in run command is not quoted, and if some users specify the path of java(like C:\java\bin\java.exe) instead of adding it to path, it will cause an error.

ouuan:
> But this feature may cause compability problem.
> In the old(and now) setting file, the program path in run command is not quoted, and if some users specify the path of java(like C:\java\bin\java.exe) instead of adding it to path, it will cause an error.

Then it's better not to change. We can wait until someone requests for this feature.

coder3101:
I was thinking how will user add breakpoint? Does QCodeEditor has it now?

ouuan:
> BTW, Your code looks Okay to me

> I was thinking how will user add breakpoint? Does QCodeEditor has it now?

So you approved these changes without trying it?

neko-para:
> I was thinking how will user add breakpoint? Does QCodeEditor has it now?

QCodeEditor doesn't suppot marks on the left margin. Currently the 'add' button will add a breakpoint according to the cursor. It needs changed, but now it's just for test. Meanwhile, the cursor in the editor will automatically change according to the current line(when running)

ouuan:
It often crashes when I press "start" in the debug window on my machine :thinking: 

Is this still work in progress? Next time you can open a draft PR for work in progress.

coder3101:
> > BTW, Your code looks Okay to me
> 
> > I was thinking how will user add breakpoint? Does QCodeEditor has it now?
> 
> So you approved these changes without trying it?

No I just meant that code quality and how it is organised looks good. Before merging we will try it for sure!

coder3101:
@neko-para are you waiting on some other feature before this can be implemented? It has been quite a long time since we got any activity here!

neko-para:
Yes, I've working on the new preferences window before.
I still don't know how to highlight a certain line(to show breakpoints and the current running row).
I'll restart updating it soon, but I need some features of QCodeEditor.

neko-para:
Maybe i can move the branch from my fork to this repository?

neko-para:
But it is a big job to make that old branch to fit the new code structure.

neko-para:
All codes will be moved to a local branch from the forked one.

neko-para:
Move to [340](https://github.com/cpeditor/cpeditor/pull/340)

Issue #204, opened 2020-03-15
add multi testcases in a time


StudyingFather:
**Is your feature request related to a problem? Please describe.**

Now the input file and output file can only be added one by one.

**Describe the solution you'd like**

See the title.

**Describe alternatives you've considered**

**Additional context**

ouuan:
In addition: we can let the user set regex patterns for a pair of input and output (like [lemon](https://github.com/Sojiv/Project_lemon)).

ouuan:
@allcontributors add @StudyingFather as a contributor for ideas.

Issue #166, opened 2020-02-09
New preference window with scrollable tabs on the left


ouuan:
**Is your feature request related to a problem? Please describe.**

The current preference window is not able to contain too many options.

**Describe the solution you'd like**

Use a preference window with scrollable tabs on the left. i.e. Make tabs be QPushButtons in a QScrollArea on the left.

And users should be able to search for options.

**Describe alternatives you've considered**

Make the default size of the preference window bigger, so more tabs can be placed in it without the left and right buttons.

**Additional context**

N/A

coder3101:
As of now, I have increased the size of the preference window when it is opened and so it can contain most of the preference tabs directly. As for searching and new preference window UI, It can be under-prioritized now.

coder3101:
637d6b8

ouuan:
I'll write a completely new preference window.

ouuan:
The first-step plan:

![first-step plan](https://user-images.githubusercontent.com/30581822/77845344-c29e7900-71e0-11ea-8062-07d60a00b1c1.png)

The near-future plan:

![near-future plan](https://user-images.githubusercontent.com/30581822/77845345-c500d300-71e0-11ea-99c0-a449c48d88ff.png)

Feel free to share your opinions on these plans.

ouuan:
Other settings pages except for the snippets (which is complicated) and the Limits (which is a new feature) are easy to implement, especially the Actions section (which has only checkboxes). Contributions are welcomed!

PR #246, opened 2020-04-22
Add multi testcases

neko-para:
## Description
Add multi testcases in a time.

## Related Issue
\#204

## Type of changes

### DIFF
--- a/src/Settings/PreferencesWindow.cpp
+++ b/src/Settings/PreferencesWindow.cpp
@@ -129,6 +129,8 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
     addPage("Advanced/Limits",
             {"Time Limit", "Output Length Limit", "Message Length Limit", "HTML Diff Viewer Length Limit",
              "Open File Length Limit", "Load Test Case File Length Limit"});
+
+    addPage("Test", {"Testcase Scan Rule"});
 }
 
 void PreferencesWindow::display()
--- a/src/Settings/PreferencesWindow.cpp
+++ b/src/Settings/PreferencesWindow.cpp
@@ -121,7 +121,7 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
 
     addPage("Extensions/CF Tool", {"CF/Path"});
 
-    addPage("File Path/Testcases", {"Add Testcases From Files Rule"});
+    addPage("File Path/Testcases", {"Testcases Matching Rules"});
 
     addPage("Key Bindings", {"Hotkey/Compile", "Hotkey/Run", "Hotkey/Compile Run", "Hotkey/Format", "Hotkey/Kill",
                              "Hotkey/Change View Mode", "Hotkey/Snippets"});
--- a/src/Settings/PreferencesWindow.cpp
+++ b/src/Settings/PreferencesWindow.cpp
@@ -121,6 +121,8 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
 
     addPage("Extensions/CF Tool", {"CF/Path"});
 
+    addPage("File Path/Testcases", {"Add Testcases From Files Rule"});
+
     addPage("Key Bindings", {"Hotkey/Compile", "Hotkey/Run", "Hotkey/Compile Run", "Hotkey/Format", "Hotkey/Kill",
                              "Hotkey/Change View Mode", "Hotkey/Snippets"});
 
@@ -129,8 +131,6 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
     addPage("Advanced/Limits",
             {"Time Limit", "Output Length Limit", "Message Length Limit", "HTML Diff Viewer Length Limit",
              "Open File Length Limit", "Load Test Case File Length Limit"});
-
-    addPage("Misc", {"Testcase Scan Rule"}); // TODO: move to the right group
 }
 
 void PreferencesWindow::display()
--- a/src/Settings/PreferencesWindow.cpp
+++ b/src/Settings/PreferencesWindow.cpp
@@ -130,7 +130,7 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
             {"Time Limit", "Output Length Limit", "Message Length Limit", "HTML Diff Viewer Length Limit",
              "Open File Length Limit", "Load Test Case File Length Limit"});
 
-    addPage("Test", {"Testcase Scan Rule"});
+    addPage("Misc", {"Testcase Scan Rule"}); // TODO: move to the right group
 }
 
 void PreferencesWindow::display()
--- a/src/Settings/settings.json
+++ b/src/Settings/settings.json
@@ -464,5 +464,11 @@
         "default": 500000,
         "param": "QList<QVariant> {2,1000000000}",
         "tip": "The maximum number of characters in a testcase file to load.\nA testcase file won't be loaded if it's too long."
+    },
+    {
+        "name": "Testcase Scan Rule",
+        "type": "QStringList",
+        "default": "QStringList {\"(.*)\\\\.in\", \"%1.ans\"}",
+        "param": "2"
     }
 ]
--- a/src/Settings/settings.json
+++ b/src/Settings/settings.json
@@ -466,9 +466,10 @@
         "tip": "The maximum number of characters in a testcase file to load.\nA testcase file won't be loaded if it's too long."
     },
     {
-        "name": "Testcase Scan Rule",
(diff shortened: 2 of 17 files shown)


coder3101:
You do need to add the new wrapper type and how to use it in Settings Readme and it's translation so that this PR will be fully complete.

ouuan:
It crashes when I click "Misc" in the menu.

coder3101:
> It crashes when I click "Misc" in the menu.

As I said, when I was adding the LSP settings, I also noticed this behaviour but then I rearranged how widgets are placed in a page (like first widget was checkbox then pathItem) it worked but if I put first (pathItem then checkbox) it failed to open and crashed. Maybe he also has same problem

ouuan:
> As I said, when I was adding the LSP settings, I also noticed this behaviour but then I rearranged how widgets are placed in a page (like first widget was checkbox then pathItem) it worked but if I put first (pathItem then checkbox) it failed to open and crashed. Maybe he also has same problem

1. I couldn't reproduce that with the LSP settings.

2. Read the code before speaking:

https://github.com/cpeditor/cpeditor/blob/37b06921529ccb2882dd7079f0fb37943efb6634/src/Settings/PreferencesWindow.cpp#L132

How can only one item have issues with ordering?

coder3101:
I don't see a reason to lie for crash on LSP settings. I did posted the same stack trace and screenshot and code to Telegram group. All I am saying is that there is an issue with how add page(..); works. It crashes (maybe randomly) when adding new settings. Otherwise how could it crash now when he has added things ideally?

ouuan:
> I don't see a reason to lie for crash on LSP settings.

Maybe you made other mistakes. I tried reordering and it worked well. If you believe that it still happens now, you can provide the full code (probably in a new branch).

> Otherwise how could it crash now when he has added things ideally?

You can try to find the reason or shut up. I'm sure that I can't change the order of a list with a single item.

ouuan:
I wasn't complaining "hey! your codes crashed!".

I was arguing against this:

> You do need to add the new wrapper type and how to use it in Settings Readme and it's translation so that this PR will be fully complete.

There are still many things besides documentation to do in this PR, and the strongest evidence is that it crashed on my machine.

ouuan:
I would try to find the reason before saying "It crashes when I click 'Misc' in the menu" if you didn't say "this PR will be fully complete".

coder3101:
https://github.com/cpeditor/cpeditor/blob/6d31687804c796963ab7f34878677e372957c9b0/src/Settings/PreferencesPageTemplate.cpp#L92

Maybe you should add a check for `widget` not being nullptr because `widgets[i]` is simply out of range.

> You can try to find the reason or shut up.

You should have a look at [CODE_OF_CONDUCT.md](https://github.com/cpeditor/cpeditor/blob/master/CODE_OF_CONDUCT.md)

ouuan:
> Maybe you should add a check for `widget` not being nullptr because `widgets[i]` is simply out of range.

That couldn't be out of range unless in some derived classes.

---

The reason for the crash is that the name of the setting is wrong.

ouuan:
The reason for the out-of-range is that the settings name is wrong and then the UI name is wrong.

ouuan:
I rebased all commits on master.

neko-para:
I've met the crash before, but eventually I found out that I use a wrong name. 
About ```// TODO: move to the right group ```, I mean it should be moved to another page instead of 'misc'. Maybe I should name this page 'ungrouped'?

ouuan:
> About `// TODO: move to the right group `, I mean it should be moved to another page instead of 'misc'. Maybe I should name this page 'ungrouped'?

I've moved it to `File Path/Testcases`, you can see the plan [here](https://github.com/cpeditor/cpeditor/issues/166#issuecomment-605607699).

ouuan:
I think we've encountered a bug of GitHub (and maybe even git). Let's keep this PR and this branch as it is now, and I'll contact GitHub.

In short:

1. The diff of 6d31687...4bebfc5 is wrong.
2. [The CI status](https://github.com/cpeditor/cpeditor/pull/246/checks?check_run_id=608369506) triggered by pull_request was on the commit 9f375b5f71fa4aeb9be24adb486d809d33d55b1a before the re-run. And @neko-para claims that he didn't make this commit.

ouuan:
Now I think the wrong diff is normal, just like the incomplete diff of merge commits, it only displays the "extra" changes except the merge.

See: https://github.com/cpeditor/cpeditor/compare/6d31687..4bebfc5 and https://github.com/cpeditor/cpeditor/compare/6d31687...4bebfc5

ouuan:
The commit 9f375b5f71fa4aeb9be24adb486d809d33d55b1a is mysterious, but we can just simply ignore it.

ouuan:
Wait, I think it's not normal again, because the [Files changed](https://github.com/cpeditor/cpeditor/pull/246/files) section is also wrong.

src/Util.cpp
```diff
@@ -85,10 +85,18 @@ bool saveFile(const QString &path, const QString &content, const QString &head,
...
-QString readFile(const QString &path, const QString &head, MessageLogger *log)
+QString readFile(const QString &path, const QString &head, MessageLogger *log, bool notExistWarning)
 {
     if (!QFile::exists(path))
+    {
+        if (notExistWarning)
+        {
+            if (log != nullptr)
+                log->warn(head, QString("The file [%1] does not exist").arg(path));
+            LOG_WARN(QString("The file [%1] does not exist").arg(path));
```
coder3101, line 96: Not required and duplicate logs will be produced log->warn("..."); always writes the message to the log file as well as message box of UI. Maybe you need to use 

```LOG_WARN_IF(log == nullptr, QString("The file [%1] does not exists).args(path));```
or an else statement before `LOG_WARN`, so that two same messages does not get logged.

File: src/Settings/settings.json [hunk 2]
```diff
@@ -464,5 +464,12 @@
         "default": 500000,
         "param": "QList<QVariant> {2,1000000000}",
         "tip": "The maximum number of characters in a testcase file to load.\nA testcase file won't be loaded if it's too long."
+    },
+    {
+        "name": "Testcases Matching Rules",
+        "type": "QList<QVariant>",
+        "default": "QList<QVariant> { QStringList {\"(.*)\\\\.in\", \"\\\\1.ans\"}, QStringList {\"(.*)\\\\.in\", \"\\\\1.out\"}}",
+        "param": "QList<QVariant> { QStringList {\"Input Regex\", \"The regular expression for the input file name\"}, QStringList {\"Answer Replace\", \"The replace expression for the answer file name. You can use \\\"\\\\1\\\" for the first captured group.\"}}",
+        "tip": "Pairs of regular expressions used when adding pairs of testcases from files. Each pair of regular expressions represents for a testcase."
```
coder3101, line 473:
```suggestion
        "tip": "Pairs of regular expressions used when adding pairs of test cases from files. Each pair of regular expressions represents a test case."
```

src/Settings/PreferencesWindow.cpp
```diff
@@ -121,6 +121,8 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
 
     addPage("Extensions/CF Tool", {"CF/Path"});
 
+    addPage("File Path/Testcases", {"Testcases Matching Rules"});
```
coder3101, line 124: This location could be better like "Testcases" -> "Testcase Load rules". "Testcases" -> "Save test cases" and other test cases related functions could be there. But this is not an issue and we need to restructure some other settings and it could be done even after a merge. 👍  Good work

src/Util.cpp
ouuan: `log->warn` is logged as information, but we need a warning in the log file.

coder3101: This change was brought by you where all Editor messages box messages are by default in INFO section of Log Files. what about duplication of Logs? Why put it to INFO? 

This is same as putting LOG_INFO and LOG_WARN both inside messagelogger.warn()? Also this is Util function that can be called multiple times so it can SPAM the Log file with duplicate logs.

A possible alternative will be to add a prefix "MessageLogger" before all LOGS that are written with log.info/warn/error. So,

```cpp

log->warn("Something is missing");

```

Internally writes:

`LOG_WARN("[MessageLogger]" << "Something is missing ");`

ouuan: Because not all warnings for the user should be a warning in the log file, for example, the compilation warnings.

ouuan: > Also this is Util function that can be called multiple times so it can SPAM the Log file with duplicate logs.

1. Now this is used only here (when adding paris test cases from files).
2. This is logged only when the file doesn't exist unexpectedly.

So the double logging here won't spam.

coder3101: Okay then agreed!

coder3101:
Optional thing to do should be that this new feature should be added to it's Changelog.md

PR #249, opened 2020-04-23
Misspell fixer

ouuan:
## Description

Add a workflow to fix typos.

## Motivation and Context

There are some typos in the code.

## How Has This Been Tested?

#247

## Type of changes

### DIFF
--- a/.github/workflows/misspell-fixer.yml
+++ b/.github/workflows/misspell-fixer.yml
@@ -1,0 +1,16 @@
+name: Misspell fixer
+
+on:
+  - push
+
+jobs:
+  misspell-fixer:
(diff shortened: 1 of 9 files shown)


.github/workflows/misspell-fixer.yml
```diff
@@ -0,0 +1,16 @@
+name: Misspell fixer
```
coder3101, line 1: Just a general question: Does this spell checker action guarantees that compilation will not fail? As I see it renames function names as well. Or does it only corrects spelling inside quotes?

ouuan: I don't know, but if it fails, we can revert it or fix it after merge.

PR #251, opened 2020-04-23
ci(check_format): Auto create PR to fix format

ouuan:
##  Description

Auto create PR for Clang Format check.

## Motivation and Context

In this way, we can fix wrong format more easily.

This is better than commit directly, because:

1. It works on protected branches.
2. We can choose to merge it or change it manually.
3. After the merge, CI will run on it. Though it won't before the merge, we can still revert/fix it easily after the merge.

## Type of changes

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -16,3 +16,13 @@ jobs:
       run: clang-format-9 -i src/*.cpp src/*/*.cpp src/*.hpp src/*/*.hpp
     - name: Check diff
       run: git diff --exit-code HEAD
+    - name: Create Pull Request
+      if: failure()
+      uses: peter-evans/create-pull-request@v2
+      with:
+        commit-message: "style: Format codes"
+        title: "Format codes for ${{ github.ref }}"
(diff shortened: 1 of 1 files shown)


coder3101:
Does it creates PR for PR that fails code formatting? I don't see a restriction to run this only on `push`

ouuan:
> Does it creates PR for PR that fails code formatting? I don't see a restriction to run this only on `push`

Pull requests can't use secrets, so it will fail to open a PR when it's triggered by a PR.

See #250 and [the check run](https://github.com/cpeditor/cpeditor/pull/250/checks?check_run_id=612914390).

coder3101:
So now CI will fail if a typo is in PR or its code format is not as per our styles. As with code format the PR proposer can know that he/she needs to run clang format. As for spell fixer, how will they know where is the spelling mistake?

ouuan:
> As for spell fixer, how will they know where is the spelling mistake?

1. They can enable GitHub Actions in their forks, and possibly re-push.
2. They can read the diff in the check run.

PR #268, opened 2020-04-29
feat: Handle signals

### LINKED ISSUE #178: Catch system signals like SIGINT

It terminates without saving status for hot-exit when I start it from terminal and press Ctrl+C.

Catch system signals like SIGINT and exit safely.

N/A

**Additional context**

References:
1. <https://doc.qt.io/qt-5/unix-signals.html>
2. <https://stackoverflow.com/questions/7581343/how-to-catch-ctrlc-on-windows-and-linux-with-qt>

ouuan:
## Description

Catch SIGINT and SIGTERM, normally close CP Editor when receiving these signals.

## Related Issue

This closes #178.

## Motivation and Context

1. Make it easier to start CP Editor from the terminal.
2. If hot-exit is enabled, the tab statuses are saved in some unexpected situations.

## How Has This Been Tested?

On Manjaro Linux and Windows.

It works well on Manjaro.

On Windows, if it's launched from cmd, it won't hang up; if it's launched from git bash, it sometimes doesn't work. However, it works well when killing from the task manager.

## Type of changes

## Checklist


coder3101:
Annotation 2020-04-29 131832

When the application closes normally this happens. You might not have noticed it because this happens during `quit` but VS Debugger caught it.

coder3101:
```cpp
Daemon::Daemon(QObject *parent) : QObject(parent)
{
#ifdef Q_OS_WIN32
    daemon = this;
#else
    if (::socketpair(AF_UNIX, SOCK_STREAM, 0, sigFd))
        qFatal("Couldn't create socketpair");

    notifier = new QSocketNotifier(sigFd[1], QSocketNotifier::Read, this);
    connect(notifier, SIGNAL(activated(int)), this, SLOT(handleSignal()));
#endif
}

Daemon::~Daemon()
{
    delete notifier;
}
```

If notifier is only created in UNIX why is it deleted in both OS?

coder3101:
Launching application and doing unsaved work then killing it from Task manager also does not saves the state before exit. Also, make sure if Settings are saved when the signal is activated.

It would also be a nice option to forcefully hot exit in case of signal and in the next launch ask the user to restore the session just like in other applications. If hot exit was enabled there is no need to ask, directly restore the status.

You could ask in a dialog like "Last CP Editor session was unexpectedly closed. Would you like to restore your sessions?" (Yes | No)

ouuan:
> Launching application and doing unsaved work then killing it from Task manager also does not saves the state before exit.

In the task manager, it's like this: (I can't remember exactly)

```
> CP Editor
   - a.cpp - CP Editor
```

When you kill the outer one, it will be killed. I guess in this case something like SIGKILL is sent.

When you kill the inner one, the status is saved, at least on my Windows.

ouuan:
> Finds potential security problems in read calls
Check buffer boundaries if used in a loop including recursive loops (CWE-120, CWE-20)

I have no idea how to fix this :thinking: 

I've tried to follow [this](https://stackoverflow.com/questions/53020336/read-check-buffer-boundaries-if-used-in-a-loop-including-recursive-loops), but the warning is still there. I also found [this](https://stackoverflow.com/questions/59293533/a-flaw-reported-by-flawfinder-but-i-dont-think-it-makes-sense). I think we can ignore it?

coder3101:
What is the need of `read` syscall in signal handling of UNIX. See there is are no examples that handles signals with read syscall. 

Check this: https://gist.github.com/aspyct/3462238
It will be much cleaner to implement handling this way.

ouuan:
I guess he copied from <https://doc.qt.io/qt-5/unix-signals.html> :thinking:

ouuan:
> Also, make sure if Settings are saved when the signal is activated.

Do you mean the current settings or the unsaved settings in the preferences window?

ouuan:
[This](https://stackoverflow.com/a/7582555/12601364) looks good :thinking:

ouuan:
Now it looks much better. I haven't tested on Windows yet.

ouuan:
Just tested on Windows, it works well with the task manager and `taskkill` but not with git bash when it waits for the user's input.

ouuan:
A bit problem is how to deal with QDialogs.

ouuan:
I tried many ways to close the AppWindow when there's a dialog (the hardest one to deal with is probably the progress bar when loading hot exit status). Some of them can't really close the application, some of them cause a bus error.

AFAIK, I think we have these choices:

1. Don't close when there's a dialog.
2. Allow runtime errors when closing the application with dialogs.
3. Deal all dialogs separately and manually. That will be a big work.

ouuan:
Now I only met problems in some very rare cases which can be considered never happen in daily use.

coder3101:
I will approve this PR if it works on Mac as well. It will take some time before I get to test on it. So expect some delays in review! (ETA:15 hours)

ouuan:
@coder3101 When will you review this?

coder3101:
It has been tested on Mac by @Razdeep and you should have tested it on Linux and I will test on Windows and review it.

coder3101:
It is still not working on my Windows.

The things are that the `handleSignal()` function is called in some other non-main Thread and access to any object owned by `mainThread` denies any request from other threads. So `appwindow->close()` is throwing error because appwindow is in mainThread but calling thread `handleSignal()` is in different thread.

It is due to this reason that if you run from git bash and CTRL+C, the Runtime error is coming. In order to fix this, you should use `QMetaObject::invokeMethod(appwindow, "close")`, A safe invoker that works across threads.

I tested that doing this prevents the runtime error but still does not works in saving the file. Maybe more investigation is needed.

ouuan:
68140a7 works on my Windows 10 with Qt 5.12.7 🤔

ouuan:
@coder3101 Can you try:

1. Uninstall CP Editor if you have installed one by the installer in the releases.
2. Install [6.3.2](https://github.com/cpeditor/cpeditor/releases/download/6.3.2/cpeditor-6.3.2-x64-setup.exe).
3. Download the [Artifact](https://github.com/cpeditor/cpeditor/suites/650984982/artifacts/5588371).
4. Replace the exe file in the installed CP Editor with the Artifact.
5. Test it.

coder3101:
> 
> 
> @coder3101 Can you try:
> 
>     1. Uninstall CP Editor if you have installed one by the installer in the releases.
> 
>     2. Install [6.3.2](https://github.com/cpeditor/cpeditor/releases/download/6.3.2/cpeditor-6.3.2-x64-setup.exe).
> 
>     3. Download the [Artifact](https://github.com/cpeditor/cpeditor/suites/650984982/artifacts/5588371).
> 
>     4. Replace the exe file in the installed CP Editor with the Artifact.
> 
>     5. Test it.

Still does not work. The problem is that destructor for the `~AppWindow` is called but before it runs to complete the application is being killed, making `~Appwindow` run only half. I prove this by adding a breakpoint to first lines of `~Appwindow()`, It stops at breakpoint but when I resume or jump in/to next instruction the application returns. Strangely if the breakpoint is in the last lines of
destructor, the breakpoint is never reached.

ouuan:
> The problem is that destructor for the `~AppWindow` is called but before it runs to complete the application is being killed, making `~Appwindow` run only half.

Where is it killed? SIGINT is captured.

coder3101:
Why don't we use [this simple library](https://skycoder42.github.io/QCtrlSignals/index.html) 

Your Signal handler does not works on my computer because in Windows Signal Handling should be done on a different thread, but let's leave it all. The library is easy to setup and works fine on all OS. I suggest to add in `third_party/qctrlsignal/` but not as submodule.

ouuan:
> Why don't we use [this simple library](https://skycoder42.github.io/QCtrlSignals/index.html)

I think it's more polite to say "We can use this simple library" because I never heard of this library before and never said we shouldn't use it.

> I suggest to add in `third_party/qctrlsignal/` but not as submodule.

I think we can fork it and use the fork as a submodule, in this way we can pick the upstream changes (though the last commit is more than one year ago) more easily, and show that it's a third-party submodule more clearly. What's your main reason not to use it as a submodule?

coder3101:
Closed in favour of #327

ouuan:
@coder3101 Can you try the latest commit on the branch handle-signal?

coder3101:
It still does not work. The signal is not being caught at all.

Why still working on this branch?

ouuan:
> Why still working on this branch?

Because the implementation on this branch is simpler, and I haven't figured out how to catch SIGHUP with QCtrlSignals yet.

I also want to know why is QCtrlSignals working but not this, and why is this working on my Windows 10 but not on yours.

ouuan:
> The signal is not being caught at all.

Instead of crash, the signal isn't caught? How did you find that, by adding debug statements?

coder3101:
> 
> 
> > The signal is not being caught at all.
> 
> Instead of a crash, the signal isn't caught? How did you find that, by adding debug statements?

Started from git bash and then ctrl+c did close the editor but not saved the editor status. Does that not mean signal was not caught. Doing the same on the other branch saves the settings.

ouuan:
> Started from git bash and then ctrl+c did close the editor but not saved the editor status. Does that not mean signal was not caught.

Maybe the signal was not caught, maybe the signal was caught but the status is not saved.

ouuan:
Just tested on Windows, on my machine, it works fine when killing from the task manager, and crashes (`echo $?` prints `127`) when pressing Ctrl+C in git bash.

ouuan:
#327 also exits with 127 but the status is saved 🤔

ouuan:
Maybe we can use #268 on UNIX and #327 on Windows? Or we have to either work on catching SIGHUP in #327 or catching Ctrl+C in git bash on Windows in #268.

ouuan:
@coder3101 Is this branch working on your machine when killing from the task manager?

coder3101:
> 
> 
> @coder3101 Is this branch working on your Windows when killing from the task manager?

Nope, but I think it is a problem at my side on windows, sometimes the other signal handler too does not works and sometimes works. If this works on your PC and Qt versions, I will approve this changes.

ouuan:
@neko-para Can you confirm if this works on your Windows or not?

coder3101:
image

Ctrl+C on my Arch linux is getting Seg fault?

coder3101:
A fix for this issue in free has been published in #354

ouuan:
@neko-para Can you test this on your Windows and approve it?

ouuan:
Because there's no response from @neko-para and @StudyingFather has confirmed that this is working on his Windows, so @coder3101 you can approve and merge this.

ouuan:
I finally discovered that it works on my Windows with `--config Release` but doesn't work with `--config Debug`.

coder3101:
> 
> 
> I finally discovered that it works on my Windows with `--config Release` but doesn't work with `--config Debug`.

I can confirm this too. The latest beta works fine because it is build in Release mode.

Issue #319, opened 2020-05-03
Allow user to choose the right text codec


neko-para:
> This feature is requested by @2740365712
* Issue
  Currently the file is always saved in utf-8, but some editors(like devc++) load the source according to the locale(gbk if Chinese, shiftjis if Japanese, etc.), which causes problems.

* Possible solution
  * Allow user to choose the saving codec.
  * Add a setting about the default saving codec.

* Some pictures
![_@73FC QYA8BHC}A~ DED0Q](https://user-images.githubusercontent.com/26452623/80916160-964ebd00-8d89-11ea-801b-1adcbe17f2d4.png)

neko-para:
I think the second method may be a good choice.
editor should save the file in the system locale by default, and could be changed to utf8 in setting.

PR #321, opened 2020-05-04
feat: Support text codec

neko-para:
## Description
Support basic guessing codec.
Support changing saving codec.

## Related Issue
#319

## Type of changes

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -56,6 +56,8 @@ bool saveFile(const QString &path, const QString &content, const QString &head, 
         LOG_ERR_IF(!QDir().mkpath(dirPath), QString("Failed to create the directory [%1]").arg(dirPath));
     }
     QString name = SettingsHelper::getSavingCodec();
+    if (name == "system")
+        name = QTextCodec::codecForLocale()->name();
     QTextCodec *codec = QTextCodec::codecForName(name.toUtf8());
     if (!codec)
         codec = QTextCodec::codecForName("UTF-8");
@@ -131,9 +133,12 @@ QString readFile(const QString &path, const QString &head, MessageLogger *log, b
     }
     QByteArray data = file.readAll();
     QString content;
-    for (auto name : SettingsHelper::getCodecGuessPriority())
+    for (QVariant nameVariant : SettingsHelper::getCodecGuessPriority())
     {
-        auto codec = QTextCodec::codecForName(name.toString().toUtf8());
+        QString name = nameVariant.toString();
+        if (name == "system")
+            name = QTextCodec::codecForLocale()->name();
+        auto codec = QTextCodec::codecForName(name.toUtf8());
         if (codec)
         {
             QTextCodec::ConverterState state;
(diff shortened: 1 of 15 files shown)


neko-para:
Actually state.invalidChars doesn't work perfectly.
Currently, we can put system and specified codec(like gb2312) before utf-8, as these codecs are generally more strict.
Therefore, a better solution is needed.

ouuan:
> Actually state.invalidChars doesn't work perfectly.
> Currently, we can put system and specified codec(like gb2312) before utf-8, as these codecs are generally more strict.
> Therefore, a better solution is needed.

In this case, I'll convert this PR to a draft PR, and you can open it for review when it's ready.

neko-para:
I've got up an idea. We can add a status bar which contains the info and buttons like vscode, and the codepage and line ending could also be added there.

src/Util.cpp
```diff
@@ -105,20 +112,17 @@ QString readFile(const QString &path, const QString &head, MessageLogger *log, b
                 log->warn(head, QString("The file [%1] does not exist").arg(path));
             LOG_WARN(QString("The file [%1] does not exist").arg(path));
         }
-        return QString();
+        return "";
```
ouuan, line 115: I think you didn't read this?

https://github.com/cpeditor/cpeditor/blob/6565bb5c6fa5369b2b8528785ff35bc96b4073b0/src/Util.hpp#L54-L56

There seems to be no specific reason for changing this line.

src/Util.hpp
```diff
@@ -18,6 +18,7 @@
 #ifndef UTIL_HPP
 #define UTIL_HPP
 
+#include <QByteArray>
```
ouuan, line 21:
```suggestion
```

It compiles without this include on my machine.

ouuan:
In fact, I don't think this feature is very useful :thinking: In #319, I'd rather suggest @2740365712 not to write comments in Chinese, or not to open a file with Chinese comments with both CP Editor and Dev C++. However, I'll accept your PR if there are no problems.

src/Util.cpp
neko-para: I noticed. But I think that QByteArray is the file content, and we convert it to QString by QString::fromUtf8, so I add the fromUtf8 or codec->toUnicode() to all the places used this function

neko-para:
I know why you crashes. I'll fix it in next commit.

src/Util.cpp
neko-para: And as the return type is QByteArray, it's better to return a "" than a QString()

ouuan: Is a non-exist file distinguished from an empty file now?

ouuan: Maybe the correct choice is `QByteArray()`?

ouuan: And you should also modify the comments.

neko-para: ok, I'll do it

PR #327, opened 2020-05-07
feat: Gracefully exit on Signals 

coder3101:
## Description
Adds Signal handler with help of QCtrlControl 

## Type of changes

## Checklist

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "third_party/testlib"]
 	path = third_party/testlib
 	url = https://github.com/MikeMirzayanov/testlib
+[submodule "third_party/QCtrlSignals"]
+	path = third_party/QCtrlSignals
+	url = https://github.com/cpeditor/QCtrlSignals
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ add_subdirectory(third_party/QtFindReplaceDialog/dialogs)
 
 add_subdirectory(third_party/lsp-cpp)
 
+add_subdirectory(third_party/QCtrlSignals)
+
 IF(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
   FIND_PACKAGE(Git)
   IF(GIT_FOUND)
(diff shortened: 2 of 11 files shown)


coder3101:
I have left the handing signal with a comment on what should be expected behaviour or what could be better than this as per your wish. I have ported the changes to appwindow and other helper API from #268 into this.

I also confirm that now, I am able to exit gracefully (with hot exit enabled) and that this library works fine on my windows.

ouuan:
Now it doesn't catch SIGHUP, which is sent on Linux when closing the terminal where the application was started :thinking:. However, I saw SIGHUP in `qctrlsignalhandler_unix.cpp`, we may need to find out how to catch it later.

And, without this:

https://github.com/cpeditor/cpeditor/blob/68140a7903b700b443063a721a5e1e5ae3bc2edd/src/Application.cpp#L255

, the application won't be killed after receiving SIGTERM during the progress dialog.

ouuan:
Please test the latest commit first.

coder3101:
> Please test the latest commit first.

It still works on my Windows but when Hotexit is not enabled `SIGINT` opens the dialog box to ask for saving changes, and same with killing from task manager.

If it will pop open the `closeConfirm` dialog what good it is? I thought application would exit normally and in next launch show a dialog "Unexpected close, want to restart?"

**I have not still ran this commit, will confirm this behaviour in 5-6 hours, I tested locally by putting exact code as you have for signal handling now, (copy pasted old handling code) and found this behaviour**

ouuan:
> If it will pop open the `closeConfirm` dialog what good it is? I thought application would exit normally and in next launch show a dialog "Unexpected close, want to restart?"

I think this is useful on Linux. SIGINT is the same as a normal close, so the user can use Ctrl+C to normally close it. It is strange that the dialog is also showed when it's killed from the Windows task manager, maybe we can treat SIGINT as SIGTERM on Windows.

ouuan:
I just noticed that d5ea61ba4d6c5fe553c32d877ea1de14529883b2 has been not working since the new preferences window, because now the settings are saved when normally exiting. After this PR is finished and merged, the feature introduced by that commit will be more useless, so instead of making it work again, we can simply remove it.

coder3101:
What behaviour you want when SIGTERM and SIGINT is invoked in Windows( from git bash)? I will try to fix it tell me behaviour with Hotexit and without Hotexit?

> Note that killing from Task manager sends SIGINT

You work on behaviour for UNIX.

PR #357, opened 2020-05-16
Fix #271 and #353

### LINKED ISSUE #271: Add an option to save the executable files in the temp directory or the source file directory
Automatically delete the filename.out file after execution or on closing the tab for that problem in editor.

As we click to submit other problems on codeforces, the folder contains the .out files of several problems along with the .cpp files.
These files should be deleted to make the folder cleaner.

having a common a.out file instead of filename.out or deletion of the .out file after exectution. Also tell if there are any arguments that can be passed to fix this.
we can have something like rm filename.out
**screenshot**

Screenshot from 2020-04-29 18-39-38
The problems candies, constant palindrome sum have been done on CP editor and have those .out files there whereas others done using sublime have them removed using custom build

### LINKED ISSUE #353: Open File Length Limit causes file loss after language server is added
**Describe the bug**

The origin file will be lost if the Open File Length Limit is exceeded. In v6.3, only the editor will be changed to "Open File Length Limit Exceeded", but it won't be saved to the file.

**To Reproduce**

Open a file that is longer than the Open File Length Limit.

**Screenshots**

image

**Environment**

 - OS: Manjaro KDE
 - CP Editor Version: since 5fc3dd9.

**Additional context**

I'm investigating this.

And it's better to change this behavior even before the Language Server is added, because the user may save the text "Open File Length Limit Exceeded" by accident.

ouuan:
### Added

- Now it's optional to save the file on compilation and execution or not.
- Now you can choose the path of the executable file of C++. (#271)

### Fixed

- Fixed some unexpected file saving. (#353)

### Changed

- Open an empty untitled tab when the open file length limit is exceeded. (#353)

## Related Issue

Close #271, fix #353.

## Motivation and Context

See #353.

## How Has This Been Tested?

On Manjaro KDE, tested on some basic usage.

### DIFF
--- a/src/Core/Compiler.hpp
+++ b/src/Core/Compiler.hpp
@@ -69,12 +69,14 @@ class Compiler : public QObject
     static bool check(const QString &compileCommand);
 
     /**
-     * @brief get the output path (executable file path) when compiling C++
+     * @brief get the output path (executable file path when compiling C++, class path when compiling Java)
      * @param tmpFilePath the path to the temporary file which is compiled
      * @param sourceFilePath the path to the original source file, if it's empty, tmpFilePath will be used instead of it
-     * @note the parent directory of the result will be created
+     * @param lang the language being compiled
+     * @note if lang is C++, the parent directory of the result will be created; if lang is Java, the result directory
+     * will be created
      */
-    static QString cppOutputPath(const QString &tmpFilePath, const QString &sourceFilePath);
+    static QString outputPath(const QString &tmpFilePath, const QString &sourceFilePath, const QString &lang);
 
   signals:
     /**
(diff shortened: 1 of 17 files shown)


src/Core/Compiler.hpp
```diff
@@ -66,6 +68,14 @@ class Compiler : public QObject
      */
     static bool check(const QString &compileCommand);
 
+    /**
+     * @brief get the output path (executable file path) when compiling C++
+     * @param tmpFilePath the path to the temporary file which is compiled
+     * @param sourceFilePath the path to the original source file, if it's empty, tmpFilePath will be used instead of it
+     * @note the parent directory of the result will be created
+     */
+    static QString cppOutputPath(const QString &tmpFilePath, const QString &sourceFilePath);
```
coder3101, line 77: Can you make this function more generic. By returning Exe path, Compiled Java class and "" (filename) for python scripts. Maybe you can change it to `outputPath()`. You will also add a setting for File path for Java solutions class

PR #370, opened 2020-05-23
feat: New App Updater Dialog Window

### LINKED ISSUE #361: open a window when checking for updates

When I click "check for updates", all I can do is to wait for results.

If I can not connect with the Github server, I can not get any response about this situation.

1. Open a window when checking for updates so that I know the update checker is working. And if possible I can close the window to terminate the update check.
2. Add a timeout. If the checker do not get response within the timeout, show the error message.

**Additional context**

coder3101:
## Description

`Telemetry/UpdateChecker` only checks for new updates and sends a signal
with information wheather an update is available or not. It no longer interacts with UI components directly instead widgets and custom
dialogs will show changelogs and direct download button depending upon OS for release binaries.

It also provides some finer control over the update checker, so user can
see indeterminant progress of the update check and cancel the
check if it takes too much time. The Update UI has been made much cleaner and so has been the
code for updater.

The changelog view supports full markdown formating, with the exeception
that mardown literals like `#` are also shown.

## Related Issue
#361 

## Motivation and Context
NA

## How Has This Been Tested?
Windows 10

image
image

## Type of changes

## Checklist


src/Widgets/UpdatePresenter.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    information = meta;
+
+    textEdit->setText(information.body);
+
+    delete downloadButton;
+    downloadButton = new QPushButton("Download", this);
+    subLayout->insertWidget(0, downloadButton);
+
+    // Instead of delete and re create, if we can somehow disconnect the old connection and set new one, it would be
```
coder3101, line 64: @ouuan If you know how to disconnect signal and connect a new lambda, please ammend this!

coder3101:
Codacy is reporting false positives. Hopefully, you can check it out.

ouuan:
I roughly went through the codes and made these changes. It's hard to list/explain all these changes, so I won't, and you can ask me with inline comments if you have any questions about these changes. I haven't finished reviewing yet, just making a commit first to make things clearer and review the whole PR later.

ouuan:
I think it's good to list more than the latest changelog, for example, all changelogs from the next version of the current version to the latest version.

However, if we are sticking to the current versioning rule and will never change it in the future, it's better to list only the first version of each minor version after the current version, and all versions of the latest version (if the latest version is of the same minor version as the current version, then only the new patch versions). We may also add some checks to determine whether the versioning rule has changed, though I'm not sure how to do that. It's also OK to ignore the possibility that we change the versioning rule in the future because it doesn't harm much if the displayed changelog is incomplete or with extra information.

coder3101:
I also think the current changelog text is not very sensible. Suppose, I am using 6.6.0 (which has current updater) and I haven't opened editor for a long time to the point that latest 6.8.1 comes with feature XXX and it is followed by 6.8.2 which fixes a critical issue with XXX. Naturally while releasing the update for 6.8.2 we will put in its changelog 

> Same as 6.8.1 except that a issue is fixed where XYZ occurred when doing MNQ

Now, when I launch editor after a long time today, I will get the that update is available but changelog will only say "Same as 6.8.1...", this isn't very clear. User can not know these changes without visiting GitHub release section.


A much better approach here will be to show all changelog since 6.6.0 to 6.8.2, i.e My version to Latest version. This can be done very easily, in update checker if we instead of breaking out of release iterating, continue to iterate and append the changelog. Unless we reach, 6.6.0 (current user version) or 6.5.0 ( Lower limit since this version added new updater) 

When in future if we change versioning rule, we have to make sure at least old updaters are able report new updates. Changelog is not very important since we will not change versioning rule so often.

ouuan:
> I also think the current changelog text is not very sensible. Suppose, I am using 6.6.0 (which has current updater) and I haven't opened editor for a long time to the point that latest 6.8.1 comes with feature XXX and it is followed by 6.8.2 which fixes a critical issue with XXX. Naturally while releasing the update for 6.8.2 we will put in its changelog
> 
> > Same as 6.8.1 except that a issue is fixed where XYZ occurred when doing MNQ
> 
> Now, when I launch editor after a long time today, I will get the that update is available but changelog will only say "Same as 6.8.1...", this isn't very clear. User can not know these changes without visiting GitHub release section.
> 
> A much better approach here will be to show all changelog since 6.6.0 to 6.8.2, i.e My version to Latest version. This can be done very easily, in update checker if we instead of breaking out of release iterating, continue to iterate and append the changelog. Unless we reach, 6.6.0 (current user version) or 6.5.0 ( Lower limit since this version added new updater)
> 
> When in future if we change versioning rule, we have to make sure at least old updaters are able report new updates. Changelog is not very important since we will not change versioning rule so often.

Basically, you are saying "yes, let's use your second plan, 'list only the first version of each minor ... new patch versions)'"?

ouuan:
Or the first plan?

Your reply is too long, and I can see a lot of duplicated statements of my comment above.

ouuan:
Basically, I don't think you need to use a long statement to agree with my plan while there's no one against it.

coder3101:
@ouuan Not to confuse with example I presented above, I want "show all changelog starting user version to Latest version" regardless of major, minor or patch. 


This way we can make sure that end user got a complete changelog (maybe big) but complete.

By your plan what changelog range you want if user is checking for update ( v7.2.1) while using 6.8.2??

I suggest a complete range (6.8.2-7.2.1) changelog of all release in the range.

ouuan:
> This way we can make sure that end user got a complete changelog (maybe big) but complete.

Yes, but also with many duplicates.

> By your plan what changelog range you want if user is checking for update ( v7.2.1) while using 6.8.2??

A better example is from 6.5.2 to 6.10.3: 6.6.1, 6.7.1, 6.8.1, 6.9.1, 6.10.1, 6.10.2, 6.10.3.

ouuan:
I think we can change the patch version 1 to the first patch version of a minor version, so that even if we start the release of a minor version by patch version 0, it still works.

ouuan:
BTW, now the first stable release contains the changelog of the beta releases of the same minor version, so if we use the releases API to get the changelog, we need to use only the changelogs of the 
stable versions except for the latest minor version if the latest version is beta.

ouuan:
In fact, with the current versioning rule, the best solution is to use the changelog in the version branch, from the next minor version (if the current version and the latest version are not of the same minor version)/the next patch version (if the current version and the latest version are of the same minor version) to the latest version.

coder3101:
From my understanding, here is what I have understood.

- [6.7.7 - 6.7.12] : 6.7.8, 6.7.9, 6.7.10, 6.7.11, 6.7.12 (Change in Patch)
- [6.9.0 - 7.1.1]: 7.0.0, 7.0.1, 7.0.2, 7.1.1 (Change in Major)
- [6.6.1 - 6.8.2]: 6.6.2, 6.7.1, 6.7.2, 6.8.1, 6.8.2 (Change in Minor)

Please let me know some more examples, I will work on this. Please correct me or present complete set of examples to demonstrate the behaviour.

ouuan:
> From my understanding, here is what I have understood.
> 
> * [6.7.7 - 6.7.12] : 6.7.8, 6.7.9, 6.7.10, 6.7.11, 6.7.12 (Change in Patch)
> * [6.9.0 - 7.1.1]: 7.0.0, 7.0.1, 7.0.2, 7.1.1 (Change in Major)
> * [6.6.1 - 6.8.2]: 6.6.2, 6.7.1, 6.7.2, 6.8.1, 6.8.2 (Change in Minor)

What did you understand? My first plan ("for example, all changelogs from the next version of the current version to the latest version" instead of the next paragraph "However, it's better")? Or your own plan?

My suggestion is "only the first version of each minor version after the current version, and all versions of the latest minor version (if the latest version is of the same minor version as the current version, then only the new patch versions)".

About your "understanding":

1. 7.0 is not necessarily right after 6.9, there can be 6.10, 6.11...
2. The patch version 0 is for the alpha version (master) and has no changelog.
3. This is "all versions from the next version of the installed version to the latest version" instead of my suggestion.
4. My example: [6.5.2 - 6.10.3] (no matter each minor version has how many patch versions): 6.6.1, 6.7.1, 6.8.1, 6.9.1, 6.10.1, 6.10.2, 6.10.3.
5. You can read the changelogs and see how many duplicates there will be, if all changelogs from the next version of the installed version to the latest version are shown.
6. For your examples:
    - [6.7.7 - 6.7.12] : 6.7.8, 6.7.9, 6.7.10, 6.7.11, 6.7.12 (Change in Patch)
    - [6.9.0 - 7.1.1]: 6.10.1 (if exists), 6.11.1 (if exists), ..., 7.0.1, 7.1.1 (Change in Major). For this exmaple, 6.9.1 may also be there, but it's not necessary, because there's only one case for this: a user is using the alpha version, but he/she only checks for stable releases.
    - [6.6.1 - 6.8.2]: 6.7.1, 6.8.1, 6.8.2 (Change in Minor)

coder3101:
Choose a behaviour in which at last minute of start of contest the Editor doesn't forcibly asks user to update application.

ouuan:
> Choose a behaviour in which at last minute of start of contest the Editor doesn't forcibly asks user to update application.

What does this mean? Which behavior forces the user the update? What are the options to "choose" from?

coder3101:
@ouuan If update diaalog always remains on top, how will user code? Suppose they opened editor 2 min prior to contest and "startup update check" presented the dialog that shows update available and remains on top. That is not a behaviour we want.

ouuan:
> @ouuan If update diaalog always remains on top, how will user code? Suppose they opened editor 2 min prior to contest and "startup update check" presented the dialog that shows update available and remains on top. That is not a behaviour we want.

`Util::showWidgetOnTop` is to show a widget on top, no matter they were minimized, hidden by other windows or closed. It's not to make the window a modal window.

ouuan:
Do you think there's a better name for `Util::showWidgetOnTop`? Maybe that name is somehow misleading.

coder3101:
It's okay

ouuan:
Do you understand and agree with my changelog plan now? I'm willing to work on it.

coder3101:
> Do you understand and agree with my changelog plan now? I'm willing to work on it.

Yes, I agree. Your opinion counts more on this since you have devised our release plan and you create the releases.

coder3101:
If this PR is ready, Can we have an Approving review now.

ouuan:
> If this PR is ready, Can we have an Approving review now?

I was waiting for your review :thinking: (Because this is your PR, I mean a review comment instead of an actual review).

coder3101:
I checked it and it works, So I approve it with this comment.

PR #404, opened 2020-06-13
feat: Add support for styles and themes in Settings

### LINKED ISSUE #265: Support app-wide dark theme in the settings

Now, we cannot use app-wide dark theme on GNOME desktop environment.

Support app-wide dark theme on GNOME.

~~When launching, force the user to switch to KDE~~

**Additional context**

Currently using Ubuntu 20.04 LTS with default GNOME

coder3101:
## Description

Let user change choose any style and one of two styles. Dark Fusion and Light Fusion. On windows there is another style called Auto Fusion that toggles between Dark and Light Fusion depending upon the System Settings.

Now user no longer needs to manage the System Settings to change themes. It can be done from the Preferences of the Editor. As a result of it, on GTK (GNOME), the application can use Dark theme (user can explicitly turn it on).

## Related Issue
Fixes #265 

## Motivation and Context
See #265 

## Screen Shots
- Windows  Light Fusion
  
image

 - Windows Dark Fusion

image

 - Windows Auto Fusion

 Toggles between two Fusion Styles depending upon the System Settings mode in Windows 10

 - Windows Vista Style (provided by QtStyleFactory::create("windowsvista")) **DEFAULT**

image
 

  - Windows Style (provided by QStyleFactory::create("Windows"))

image

 - GNOME Windows Style (provided by QStyleFactory::create("Windows"))

image

 -  GNOME Light Fusion **DEFAULT**

image

 - GNOME Dark Fusion 

image

 - macOS Windows Style with Light System Theme (provided by QStyleFactory::create("Windows"))

image

 - macOS Windows Style with Dark System Theme 

image

 - macOS macintosh Style with Light System Theme (provided by QStyleFactory::create("machintosh")) **DEFAULT**

image

 - macOS macintosh Style with Dark System Theme **DEFAULT ON SYSTEM DARK**

image

 - macOS Light Fusion with Light and Dark System Theme (provided by QStyleFactory::create("fusion"))

image

 - macOS Dark Fusion with Light and Dark System Theme

image

## How Has This Been Tested?
  - [x]  Styles Tested on Windows 10
  - [x]  Styles Tested on GNOME
  - [x]  Styles Tested on macOS
  - [ ]  Styles Tested on KDE

## Type of changes

## Checklist


coder3101:
The things left to add are 
- [ ] `Light Fusion` colour scheme and Light Fusion Tooltip (It will be simply inverted colour of Dark Fusion)
- [ ] UI in settings to choose the style (@ouuan I want you to do this as when the style changes a restart should be made or if we can change the style without restart, It will be best)
- [ ] Sync with master
- [ ] Testing

coder3101:
The base of this PR was created before the latest release was made. It is important to resolve them now!

coder3101:
@ouuan Can you please check all the styles that `Core::StyleManager::styleList()` provides on  KDE with both Light Breeze and Dark Breeze set.

Here are few points to note:
- I have made the assumption that a default theme is one of `QStyleFactory::keys()`. This is true to all the  OS I tested on, Windows default is called `windowsvista`, GNOME default is `light_fusion`, MacOS default is `macintosh`.

 - In order to set a style, change this value to a styleName. Recompile and Run
https://github.com/cpeditor/cpeditor/blob/e2b0b1ac214000a6ed74c16d34c1028a78c326d5/src/appwindow.cpp#L60

- Run with `auto_fusion` value, it will not change the theme as `auto_fusion` is only available with Windows. But, it will cause a LOG_ERR to be placed and it will have all the themes allowed.
https://github.com/cpeditor/cpeditor/blob/e2b0b1ac214000a6ed74c16d34c1028a78c326d5/src/Core/StyleManager.cpp#L42-L44

coder3101:
You can also work on to add a Style Settings dropdown. Just like Editor Theme.

I do not want to do it because we need to dynamically change its value as per response from `Core::StyleManager::styleList()`. On Windows set the default to `windowsvista`, `macintosh` on macOS and `light_fusion` on Linux (These are case sensitive)

Upon any change, You can pop-up and show the user that this style will be applied from next start.



To clear one thing, **There is no default Style on Linux** as whole because of different DE, It is just that Fusion is available to all DE, So I am choosing that as Default.

ouuan:
image

What's this function used for? Why not make `setStyle` public?

coder3101:
> What's this function used for? Why not make `setStyle` public?

Change of style needs re-start. So we only call `initiateStyle(SettingsHelper::getStyleName())` in constructor of Appwindow.

When user changes the settings of style, we will only call `SettingsHelper::setStyle()` not `setStyle()` because dynamically change of style does not sets correct style. So I have made it private, because it shouldn't be called more than onces!

src/Settings/AppearancePage.cpp
```diff
@@ -23,8 +23,9 @@
 #include <generated/SettingsInfo.hpp>
 
 AppearancePage::AppearancePage(QWidget *parent)
-    : PreferencesPageTemplate({"Locale", "Editor Theme", "Editor Font", "Test Cases Font", "Message Logger Font",
-                               "Opacity", "Show Compile And Run Only", "Display EOLN In Diff", "Extra Bottom Margin"},
+    : PreferencesPageTemplate({"Locale", "UI Style", "Editor Theme", "Editor Font", "Test Cases Font",
```
coder3101, line 26: "Application Style" is a better name.

ouuan:
On Arch Linux with KDE:

1. Light Fusion breaks when my system style is Dark Breeze.
2. Dark Fusion breaks when my system style is Light Breeze.
3. When switching from the breaking Fusion to the good Fusion, the good Fusion breaks, too.

1 and 2 happen even after restarting the application.

image

image

src/Settings/AppearancePage.cpp
ouuan: "UI Style" looks good when it's right under "UI Language". And I think "UI Style" is more specific. Why do you think "Application Style" is better?

coder3101:
Do not set default and the call `setStyle()`, In order for a style to be applied correctly, the style must be applied before any UI elements are created. Usually in `main()` or constructor of `AppWindow()`

src/Settings/AppearancePage.cpp
coder3101: Just to be consistent with "Editor Theme", "Application Style", So names are <Scope> <property>. If you expand UI Style, (User Interface Style) is sounds too generic. How about "Application UI Style"? Or "App UI Style"?

src/appwindow.cpp
```diff
@@ -1041,7 +1027,10 @@ void AppWindow::onSettingsApplied(const QString &pagePath)
     }
 
     if (pagePath.isEmpty() || pagePath == "Appearance")
+    {
         setWindowOpacity(SettingsHelper::getOpacity() / 100.0);
+        Core::StyleManager::setStyle(SettingsHelper::getUIStyle());
```
coder3101, line 1032: The reason for your style not being applied correctly the is because it is called way to late, call this inside constructor of Appwindow, where old windows dark theme code was. As I said, in dynamically Changing style, styles will break.

src/Settings/AppearancePage.cpp
ouuan: > Just to be consistent with "Editor Theme", "Application Style", So names are `<Scope> <property>`. If you expand UI Style, (User Interface Style) is sounds too generic. How about "Application UI Style"? Or "App UI Style"?

How can "UI" be more generic than "Application"?

src/appwindow.cpp
ouuan: I fixed it by setting the palette before the style.

src/Settings/AppearancePage.cpp
coder3101: UI and Application both are generic, I have no issue with either, it's just that abbreviations in Settings is not something I prefer.As you prefer.

ouuan: > abbreviations in Settings is not something I prefer

We have tooltips for the complete name.

src/appwindow.cpp
coder3101: Did you succeed in changing styles dynamically?

ouuan: > Did you succeed in changing styles dynamically?

Yes.

ouuan:
@coder3101 Do you approve this?

PR #411, opened 2020-06-17
feat: portable version

ouuan:
## Description

Add a portable version that:

1. saves the config file in the same directory as the binary file, instead of the system config directory.
2. on Windows, it doesn't need installation.

To build the portable version, you need to add `-DPORTABLE_VERSION=On` when executing `cmake`.

This PR also contains f14049099acb854328f568845335356975c7f2c6 from the v6.5 branch. However, we can still squash and merge this PR, and create a commit and merge before the next release.

## Motivation and Context

In this way, the user can put CP Editor in something like a USB disk, and doesn't need to worry about exporting and importing the configs.

## How Has This Been Tested?

I've tested the CI in ouuan/cpeditor, and the CI works well (except on Ubuntu, because we have to wait for the PPA).

Also tested the portable version on Windows, and even on wine on Linux it works.

However, I don't know about macOS, maybe we need something other than a `dmg` file for the portable version?

## Type of changes

- [x] Build (Changes that affect the build system or external dependencies)


coder3101:
Can you also create a feature branch on the website for v6.6 and add this new installation type to `Installing` section?

When releasing the new update we will merge that feature branch to `hugo`. As associated PR would be great, in future we might forget to add it to the website or miss to add something important.

.github/workflows/release.yml
```diff
@@ -0,0 +1,189 @@
+name: "Release"
+
+on:
+  push:
+    tags:
+      - "*"
```
coder3101, line 6: Any tag?? `tag` starting with `v` would better.

ouuan: We don't have tags started with "v".

.github/workflows/release.yml
```diff
@@ -0,0 +1,189 @@
...
+          mkdir build/out
+          mv build/Release/cpeditor.exe build/out
+          ../Qt/*/*/bin/windeployqt.exe build/out/cpeditor.exe --no-translations
+
+      - name: Pack to installer on Windows
+        if: startsWith(matrix.config.os, 'windows') && matrix.config.portable-option == 'Off'
+        shell: bash
+        run: |
+          mkdir build/out/Redist
+          cp "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Redist/MSVC/$VC_REDIST_VERSION/vc_redist.x64.exe" build/out/Redist
```
coder3101, line 168: Suppose this cp has failed, We will have a partial release because some binaries will be released like dmg and appimage, but this will not, what will be done then? delete the complete tag and release again? A hard work.

.github/workflows/release.yml
```diff
@@ -0,0 +1,189 @@
+name: "Release"
+
+on:
+  push:
+    tags:
+      - "*"
+
+env:
+  QT_VERSION: 5.15.0
+  VC_REDIST_VERSION: 14.26.28720
```
coder3101, line 10: Weekly or Bi-weekly Github Action updates the Windows Image and this number may change if VS has new updates. 
Wouldn't Line 127 cd not fail?

.github/workflows/release.yml
coder3101: Okay

coder3101:
Can you provide an EXE build with this script? I do not find any release assets in your fork of cpeditor.

ouuan:
> Can you provide an installer EXE build with this script? I do not find any release assets in your fork of cpeditor.

I published one just now.

.github/workflows/release.yml
ouuan: > Suppose this cp has failed, We will have a partial release because some binaries will be released like dmg and appimage, but this will not, what will be done then? delete the complete tag and release again? A hard work.

I spent more than one minute to understand what you mean, and I think this is much better:

If this `cp` fails, we will have a release in which there are only the `dmg`s and `AppImage`s but not the `exe`. When that happens, what should we do? If we delete the tag and release again, that will be hard work.

ouuan:
Is it OK to use a `dmg` file for the portable version on macOS? I know almost nothing about macOS.

coder3101:
> 
> 
> Is it OK to use a `dmg` file for the portable version on macOS? I know almost nothing about macOS.

Yes, it certainly portable. macOS user can run any dmg without installing, as long as it has all dependencies

PR #412, opened 2020-06-18
feat & fix & perf: improve auto-save

### LINKED ISSUE #392: False external file change when auto-save is enabled
**Describe the bug**

When auto-save is enabled, sometimes a false alert of external file change is shown.

**To Reproduce**

I haven't figured out how to reproduce it yet.

**Environment**

 - OS: Arch Linux with KDE
 - CP Editor Version: 6.5.1

**Additional context**

I'll try to find a way to reproduce it. Everyone who has ever met this can confirm it and it will be better if someone knows how to reproduce it.

ouuan:
## Description

1. Add auto-save interval in the settings.
2. Fix false external file changes when auto-save is enabled.

## Related Issue

This fixes #392.

## How Has This Been Tested?

On Arch Linux, the auto-save interval setting works fine. There's no false external file change, and real external file changes are successfully detected.


coder3101:
Why not we add a auto-save time interval in settings. (1 sec minimum to 60 sec maximum!)

So user can change it according to there PC specs.

coder3101:
Set auto-save to 500ms as lowest time would be a good idea. Saving files is an overhead.

ouuan:
> Set auto-save to 500ms as lowest time would be a good idea. Saving files is an overhead.

100ms is not an unreasonable value, it's just slow and not frozen on my computer. The minimum and maximum values could be wider than what seems good to us.

ouuan:
> remember main UI is responsible for saving

Not for auto-save, otherwise #392 won't happen.

ouuan:
Maybe you can test it on Windows?

coder3101:
> 
> 
> Maybe you can test it on Windows?

I am testing!

coder3101:
> Not for auto-save, otherwise #392 won't happen.

Because we use the default connection `Qt::ConnectionType::Auto`, which says:

> (Default) If the receiver lives in the thread that emits the signal, Qt::DirectConnection is used. Otherwise, Qt::QueuedConnection is used. The connection type is determined when the signal is emitted.

If however, we set a direct connection explicitly, then `onSaveTimerElapsed()` is invoked on the thread of the `QTimer`. From the docs:

> **Qt::ConnectionType::DirectConnection** : The slot is invoked immediately when the signal is emitted. The slot is executed in the signalling thread.


In our case our current case. If user PC takes longer than 100 ms to execute `onSaveTimerElapsed`, even before a slot is finished another connection has invoked the slot and there is a chance that main thread will be busy only doing auto-save.

On the contrary, If we use `DirectConnection` even if save takes more than 100ms, the QTimer thread is blocked for the duration of saving.

References: https://doc.qt.io/qt-5/qt.html#ConnectionType-enum

coder3101:
Now AutoSave only saves the current Tab. A better idea would be to make `onSettingsApplied()` Save everything except current tab (current tab will be saved by `onSaveTimerElapsed()`) when Autosave has been enabled. So that all unsaved changes in editor are saved as soon as autosave is enabled. After this as user changes tabs, autosave will only save the active tabs.

coder3101:
If you open a file in Visual Studio and the same file in CP Editor with autosave turned on, VS will warn you "File has changed Externally" every time autosave is invoked. Even if user is not change the file, autosave is continuously saving the file to disk and VS says that file has changed. (Maybe VS checks the last write date of file to find changes).

Here is how, I plan to solve it. 

When user is typing `onEditorTextChanged()` will start the autoSave timer and as soon as `onSaveTimerElapsed()` is called, we stop the autosave timer. Hence we will write to disk only when text really changed. This will increase performance even more.

coder3101:
@ouuan Do you approve this? You should test this on your machine as well.

ouuan:
I think we can make a separate signal for `NoChange`, otherwise, it's very weird.

coder3101:
> I think we can make a separate signal for `NoChange`, otherwise, it's very weird.

We can also get rid of enum Type as well then, and instead use a Boolean that tells is external or internal change was made.

coder3101:
Looks good to me and works fine.

ouuan:
One weird thing: if the user keeps typing and the interval is not very small, then the file is never auto-saved. Maybe we shouldn't reset the timer when a new modification is made.

coder3101:
> 
> 
> One weird thing: if the user keeps typing and the interval is not very small, then the file is never auto-saved. Maybe we shouldn't reset the timer when a new modification is made.

Yes, that will happen, but instead of doing that we can set the default auto save time delay to small say 500 ms. Now that it has better performance, it does not impact on small time.

ouuan:
> Yes, that will happen, but instead of doing that we can set the default auto save time delay to small say 500 ms. Now that it has better performance, it does not impact on small time.

Why do you think this is better? The saving still can cost noticeable time if the file (possibly the test cases) is big or the machine is slow.

coder3101:
> Why do you think this is better? The saving still can cost noticeable time if the file (possibly the test cases) is big or the machine is slow.

The saving now only happens once and that too if the user has not typed something for "autosave duration delay". One save that too when things are (semi) finalized or user is thinking something (so he/she stopped for say 500ms, this is the best time to save as QCodeEditor is ideal now). If we do not restart the timer, it means that the user is still typing and while in typing the save is being performed which on slow computers can be bottleneck as big test case and typing both are happening at same time.

ouuan:
I think we can make a setting for it (whether to reset the timer after a modification). Maybe the user wants to make the interval longer so that there are fewer saves.

ouuan:
And I think there can be a timer in each tab, otherwise it won't be saved after switching to a new tab.

coder3101:
> 
> 
> And I think there can be a timer in each tab, otherwise it won't be saved after switching to a new tab.

What if we save tab before switching to new TAB?

coder3101:
In favour of #438

Issue #417, opened 2020-06-19
Refactor Settings to support more types


neko-para:
**Is your feature request related to a problem? Please describe.**
To support custom language, many setting will be changed. Currently the snippets and editor status are load manually. But these could be loaded automatically, and more settings like these will be introduced to support custom language. So it's better to make the current settings manager able to handle these.

ouuan:
I think this is finished? Next time you can use "fixes/closes" instead of "related issue".

PR #419, opened 2020-06-19
refactor(Settings): add Object type and SettingsUpdater

neko-para:
## Related Issue
#417

### DIFF
--- a/src/Settings/CodeSnippetsPage.cpp
+++ b/src/Settings/CodeSnippetsPage.cpp
@@ -124,7 +124,7 @@ CodeSnippetsPage::CodeSnippetsPage(const QString &language, QWidget *parent) : P
 
 bool CodeSnippetsPage::areSettingsChanged()
 {
-    auto settingsKeys = SettingsHelper::getLanguageConfig(lang).querySnippets();
+    auto settingsKeys = SettingsHelper::getLanguageConfig(lang).getSnippets();
     auto UIkeys = snippetItem.keys();
     if (settingsKeys != UIkeys)
         return true;
@@ -149,7 +149,7 @@ void CodeSnippetsPage::makeUITheSameAsDefault()
 void CodeSnippetsPage::makeUITheSameAsSettings()
 {
     Util::applySettingsToEditor(editor, lang);
-    auto settingsKeys = SettingsHelper::getLanguageConfig(lang).querySnippets();
+    auto settingsKeys = SettingsHelper::getLanguageConfig(lang).getSnippets();
     for (auto key : settingsKeys)
     {
         if (!snippetItem.contains(key))
@@ -175,7 +175,7 @@ void CodeSnippetsPage::makeUITheSameAsSettings()
 
 void CodeSnippetsPage::makeSettingsTheSameAsUI()
 {
-    auto settingsKeys = SettingsHelper::getLanguageConfig(lang).querySnippets();
+    auto settingsKeys = SettingsHelper::getLanguageConfig(lang).getSnippets();
     for (auto key : settingsKeys)
     {
         if (!snippetItem.contains(key))
@@ -243,11 +243,11 @@ bool CodeSnippetsPage::aboutToSwitchToSnippet()
     if (currentItem == nullptr)
         return true;
 
-    if (SettingsHelper::getLanguageConfig(lang).querySnippets().contains(currentItem->text()) &&
+    if (SettingsHelper::getLanguageConfig(lang).getSnippets().contains(currentItem->text()) &&
         editor->toPlainText() == SettingsHelper::getLanguageConfig(lang).getSnippet(currentItem->text()))
         return true;
 
-    if (!SettingsHelper::getLanguageConfig(lang).querySnippets().contains(currentItem->text()) &&
+    if (!SettingsHelper::getLanguageConfig(lang).getSnippets().contains(currentItem->text()) &&
         editor->toPlainText().isEmpty())
         return true;
 
@@ -327,7 +327,7 @@ void CodeSnippetsPage::loadSnippetsFromFiles()
             break;
         }
         auto name = QFileInfo(file).completeBaseName();
-        if (snippetItem.contains(name) || SettingsHelper::getLanguageConfig(lang).querySnippets().contains(name))
+        if (snippetItem.contains(name) || SettingsHelper::getLanguageConfig(lang).getSnippets().contains(name))
             name = getNewSnippetName(name, true);
         if (name.isEmpty())
             break;
@@ -343,7 +343,7 @@ void CodeSnippetsPage::extractSnippetsToFiles()
         return;
     auto dir = QDir(dirPath);
 
-    auto names = SettingsHelper::getLanguageConfig(lang).querySnippets();
+    auto names = SettingsHelper::getLanguageConfig(lang).getSnippets();
 
     QString suffix;
     if (lang == "C++")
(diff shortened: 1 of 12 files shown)


neko-para:
I'm thinking about create a new page template for the QMap.
It may be useful for branch custom-language.

src/Settings/SettingsManager.cpp
```diff
@@ -88,47 +88,36 @@ void SettingsManager::loadSettings(const QString &path)
...
-        // load most of settings
-        for (const SettingInfo &si : settingInfo)
-        {
-            if (setting.contains(si.key()) && setting.value(si.key()).isValid())
-                set(si.name, setting.value(si.key()));
-            else
-                for (const QString &old : si.old)
-                    if (setting.contains(old) && setting.value(old).isValid())
+        std::function<void(QString, const QList<SettingInfo> &)> load;
+        load = [&](QString prefix, const QList<SettingInfo> &infos) {
```
sonulohani, line 92: I would not recommend making this section of code as lambda. Lambda is meant for small section of code which can be easier for user to understand. This code is way long to go into the lambda. I would instead make it a separate function.

src/Settings/SettingsManager.cpp
```diff
@@ -152,23 +141,26 @@ void SettingsManager::saveSettings(const QString &path)
...
-    }
-
-    // save editor status
-    for (QString key : keyStartsWith("Editor Status/"))
-    {
-        auto saveKey = key;
-        setting.setValue(saveKey.replace("Editor Status/", "editor_status/"), get(key));
-    }
+    std::function<void(QString, const QList<SettingInfo> &)> save;
+    save = [&](QString prefix, const QList<SettingInfo> &infos) {
```
sonulohani, line 145: Same as above. Move it in separate method.

src/Settings/SettingsManager.cpp
neko-para: The reason I make these code a lambda is that the code needs to call itself. It's ok to move it into a new function, but I think that these code just belong to the loadSettings and saveSettings, and put them inline don't make the code complex. I mean, the lambda aims to provide a way to call itself. So I don't think it's necessary to move them out.

sonulohani: Its very unclear to me as what do you mean by code calling itself. Do you mean recursion.

neko-para: Yes. Load the settings recursively when type is 'config'.

coder3101: Why not make a private member function?

src/Settings/SettingsManager.cpp
```diff
@@ -88,47 +89,37 @@ void SettingsManager::loadSettings(const QString &path)
...
-        QStringList langs = setting.childGroups();
-        if (langs.contains("Cpp"))
-        {
-            setting.beginGroup("Cpp");
-            QStringList keys = setting.allKeys();
-            for (const QString &key : keys)
-                set(QString("Snippets/C++/%1").arg(key), setting.value(key));
-            setting.endGroup();
-            langs.removeOne("Cpp");
-        }
```
ouuan, line 116: These are not in the `SettingsUpdater`.

src/Settings/settings.json
```diff
@@ -757,5 +640,21 @@
...
         "tip": "The style of the whole application."
+    },
+    {
+        "name": "Editor Status",
+        "type": "QMap",
+        "final": "QVariantMap"
+    },
+    {
+        "name": "Language Config",
+        "type": "config",
```
ouuan, line 651: What about "Object"?

src/Settings/CodeSnippetsPage.cpp
```diff
@@ -124,15 +124,14 @@ CodeSnippetsPage::CodeSnippetsPage(const QString &language, QWidget *parent) : P
...
     auto UIkeys = snippetItem.keys();
-    for (auto &key : UIkeys)
-        key = snippetKey(key);
     if (settingsKeys != UIkeys)
         return true;
     if (currentItem == nullptr)
         return false;
-    return editor->toPlainText() != SettingsManager::get(snippetKey(currentItem->text()));
+    return editor->toPlainText() != SettingsHelper::getLanguageConfig(lang).getSnippets(currentItem->text());
+    ;
```
ouuan, line 134:
```suggestion
```

src/Settings/settings.json
```diff
@@ -757,5 +640,21 @@
...
+        "name": "Editor Status",
+        "type": "QMap",
+        "final": "QVariantMap"
+    },
+    {
+        "name": "Language Config",
+        "type": "Object",
+        "sub": [
+            {
+                "name": "Snippets",
```
ouuan, line 654:
```suggestion
                "name": "Snippet",
```

The functions should be `querySnippet`, `removeSnippet`, etc. because they only operate on a single snippet.

neko-para: Actually querySnippets return a QStringList, so querySnippets is better. Maybe I can add an entry?

ouuan: > Actually querySnippets return a QStringList, so querySnippets is better. Maybe I can add an entry?

I didn't check that. Then I mean `getSnippets` should be `getSnippet`.

neko-para: Yes. I add a new entry sname for this. The setXXX getXXX removeXXX use the sname and queryXXX use the name.

src/Settings/settings.json
```diff
@@ -757,5 +640,22 @@
...
+                "sname": "Snippet",
```
ouuan, line 655:
```suggestion
                "name": "Snippet",
                "plural": "Snippets",
```

tools/genSettings.py
```diff
@@ -4,6 +4,90 @@
...
+        typename = t["type"]
+        if typename == "Object":
+            f.write(f"{ids}struct {key} {{\n")
+            f.write(f"{ids}    QString pre;\n")
+            f.write(f"{ids}    {key}(QString p) : pre(p) {{}}\n")
+            writeHelper(f, t["sub"], "pre, ", indent + 1)
+            f.write(f"{ids}}};\n")
+            f.write(f"{ids}inline {key} get{key}(QString key) {{ return {key}(QStringList {{{pre} {json.dumps(name)}, key}}.join('/')); }}\n")
+            f.write(f"{ids}inline void remove{key}(QString key) {{ SettingsManager::remove(SettingsManager::keyStartsWith(QStringList {{{pre} {json.dumps(name)}, key}}.join('/'))); }}\n")
+            f.write(f"{ids}inline QStringList query{key}() {{ return SettingsManager::itemUnder(QStringList {{{pre} {json.dumps(name)}}}.join('/') + '/'); }}\n")
```
ouuan, line 21:
```suggestion
            f.write(f"{ids}inline QStringList query{key}() {{ return SettingsManager::itemUnder(QStringList {{{pre} {json.dumps(name)}}}.join('/') + '/'); }}\n")
```

I think this can also be `get`, why `query`?

neko-para: queryXXX return all entrys and getXXX return certain data

ouuan: Why not `getSnippet` and `getSnippets`?

ouuan:
image

neko-para:
emmm. I think the two function is too similar and may mislead others.

neko-para:
If so, overload may be a better choice.

ouuan:
I think it's confusing to give two names, I might end up with `getSnippets` -> backspaces -> `querySnippets` or `querySnippet`-> backspaces -> `getSnippet`.

I think it will cause a compilation error if I accidentally added an `s` or missed an `s`.

neko-para:
Ok, I'll change it.

Issue #422, opened 2020-06-20
Monospace as default font even though its not installed


sonulohani:
**Describe the bug**

I am running cpeditor in windows 10. The default font looked ugly so I though of changing it. However I noticed that the default font is set to monospace even though the font is not installed.

**To Reproduce**
1. Open cpeditor.
2. Go to Options->Preferences->Appearance
3. **Monospace** font is set to **Editor font**, **Test Case Font** and **Message Logger Font** even though the monospace is not installed.  

**Screenshots**

![image](https://user-images.githubusercontent.com/2958681/85190921-9a538180-b2da-11ea-920d-46b259818c6e.png)

**Environment**

 - OS: Windows 10
 - CP Editor Version: Latest master (6.6.0)

Expected behavior : In windows, **Consolas** is the default monospace font and is what set as default in this editor as well.

sonulohani:
I can work on this issue. Please assign it to me if nobody else is gonna work.

coder3101:
`Consolas` will be good default for Windows, as it comes installed on all Windows and is a mono-spaced font.

neko-para:
@ouuan I don't know how to use allcontributors. Can you do it?

ouuan:
> @ouuan I don't know how to use allcontributors. Can you do it?

You can either read the document of all contributors or read what I commented in issues and pull requests. I did nothing privately, you can see all the operations in the comments.

coder3101:
@all-contributors add @sonulohani for bug

coder3101:
Steps to Reproduce https://github.com/cpeditor/cpeditor/issues/425

- If you have any other non-default font, proceed to next step otherwise set to some other font like "Consolas" and Hit Apply. The Font should change without size change.
- Hit Default for Appearance Settings and Click Apply
- The font text changes to "monospace 12" but the size is not 12 in the editor. A Restart fixes the size.

Reproduced on windows. Will check if reproducible on Linux!

ouuan:
I can confirm that on my Arch Linux with KDE, when resetting to default, the font is actually in the size of the system font.

PR #424, opened 2020-06-20
Default font set to consolas for windows os

sonulohani:
Default font in windows is by-default set to **monospace** which is incorrect. This PR fixes the default font to consolas

Issue: https://github.com/cpeditor/cpeditor/issues/422 https://github.com/cpeditor/cpeditor/issues/425

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -20,6 +20,7 @@ With the portable version, you can easily store it in something like a USB disk,
 ### Changed
 
 - Now the lengths of the files added by "Add Pairs of Testcases From Files" are limited by Preferences->Advanced->Limits->Load Test Case File Length Limit. (#405)
+- Default font is set to consolas for windows operating system. (#422)
 
 ## v6.5
 
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -16,6 +16,7 @@ With the portable version, you can easily store it in something like a USB disk,
 ### Fixed
 
 - Fix the wrong default setting for C++ executable file path.
+- Fix Font Size restores to small size after set to default or first launch.
 
 ### Changed
 
--- a/src/Settings/settings.json
+++ b/src/Settings/settings.json
@@ -16,7 +16,10 @@
     {
         "name": "Editor Font",
         "type": "QFont",
-        "default": "QFont(\"monospace\")",
+        "default": {
+            "Windows": "QFont(\"Consolas\")",
+            "other": "QFont(\"monospace\")"
+        },
         "old": [
             "font"
         ],
@@ -700,7 +703,10 @@
     {
         "name": "Test Cases Font",
         "type": "QFont",
-        "default": "QFont(\"monospace\")",
+        "default": {
+            "Windows": "QFont(\"Consolas\")",
+            "other": "QFont(\"monospace\")"
+        },
         "tip": "The font of test cases"
     },
     {
@@ -712,7 +718,10 @@
     {
         "name": "Message Logger Font",
         "type": "QFont",
-        "default": "QFont(\"monospace\")",
+        "default": {
+            "Windows": "QFont(\"Consolas\")",
+            "other": "QFont(\"monospace\")"
+        },
         "tip": "The font of the message logger"
     },
     {
--- a/src/Settings/settings.json
+++ b/src/Settings/settings.json
@@ -17,7 +17,7 @@
         "name": "Editor Font",
         "type": "QFont",
         "default": {
-            "Windows": "QFont(\"Consolas\")",
+            "Windows": "QFont(\"Consolas\", 12)",
             "other": "QFont(\"monospace\")"
         },
         "old": [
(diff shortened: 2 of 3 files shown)


coder3101:
Document your changes to `doc/CHANGELOG.md` under unreleased section, under subsection **Changed**.

sonulohani:
> 
> 
> Document your changes to `doc/CHANGELOG.md` under unreleased section, under subsection **Changed**.

Done.

neko-para:
@sonulohani Can you let the platform feature not restrict to QFont? Maybe other settings need this feature as well.

src/Settings/settings.json
```diff
@@ -16,7 +16,10 @@
     {
         "name": "Editor Font",
         "type": "QFont",
-        "default": "QFont(\"monospace\")",
+        "default": {
+            "Windows": "QFont(\"Consolas\")",
```
coder3101, line 20:
```suggestion
            "Windows": "QFont(\"Consolas\", 12)",
```

src/Settings/settings.json
```diff
@@ -700,7 +703,10 @@
     {
         "name": "Test Cases Font",
         "type": "QFont",
-        "default": "QFont(\"monospace\")",
+        "default": {
+            "Windows": "QFont(\"Consolas\")",
```
coder3101, line 707:
```suggestion
            "Windows": "QFont(\"Consolas\", 12)",
```

src/Settings/settings.json
```diff
@@ -712,7 +718,10 @@
     {
         "name": "Message Logger Font",
         "type": "QFont",
-        "default": "QFont(\"monospace\")",
+        "default": {
+            "Windows": "QFont(\"Consolas\")",
```
coder3101, line 722:
```suggestion
            "Windows": "QFont(\"Consolas\", 12)",
```

coder3101:
> 
> 
> @sonulohani Can you let the platform feature not restrict to QFont? Maybe other settings need this feature as well.

I had already asked for a feature like this in Settings. Platform-specific default is a nice addition.

sonulohani:
> 
> 
> We should use `QFont::TypeWriter` instead of a hardcoded font.

Hi @ouuan,

For now we have settings file as json which has hardcoded values. One thing is that we can remove default font( set "default" as empty under font section) and in code we can set font to default style hint. Please let me know your opinion.

neko-para:
@sonulohani You can use a lambda function to get the right font. The "default" is copied to source code.

sonulohani:
> 
> 
> @sonulohani Can you let the platform feature not restrict to QFont? Maybe other settings need this feature as well.

Done

sonulohani:
> 
> 
> @sonulohani You can use a lambda function to get the right font. The "default" is copied to source code.

Could you please explain.

coder3101:
> For now we have settings file as json which has hardcoded values. One thing is that we can remove default font( set "default" as empty under font section) and in code we can set font to default style hint. Please let me know your opinion.

No, settings.json does not have hardcoded values, it has codes that are expanded in the generated .hpp/.cpp files. As neko said, you can use lambda function like like:

```
default: "[]{/* write some C++ here, return the default QFont */}"

```

It should work.

sonulohani:
> 
> 
> > For now we have settings file as json which has hardcoded values. One thing is that we can remove default font( set "default" as empty under font section) and in code we can set font to default style hint. Please let me know your opinion.
> 
> No, `settings.json` does not have hardcoded values, it has codes that are expanded in the generated .hpp/.cpp files. As neko said, you can use lambda function like like:
> 
> ```
> default: "[]{/* write some C++ here, return the default QFont */}"
> ```
> 
> It should work.

Got it. Thanks.

neko-para:
@sonulohani It should be like "default": "[]() -> QFont { /* some code */ }()"
remember to call it, or the lambda will be stored and no error will appear.

doc/CHANGELOG.md
```diff
@@ -16,10 +16,12 @@ With the portable version, you can easily store it in something like a USB disk,
 ### Fixed
 
 - Fix the wrong default setting for C++ executable file path.
+- Fix Font Size restores to small size after set to default or first launch.
```
ouuan, line 19: Are you sure this is a "bug"? The font size definitely changes after setting to default. The bug is #422 and nothing else.

sonulohani: I am able to reproduce this in current master.

ouuan:
I think we can close this and make a new PR, because these changes won't be used.

sonulohani:
> I think we can close this and make a new PR, because these changes won't be used.

Okay @ouuan .

doc/CHANGELOG.md
coder3101: > Are you sure this is a "bug"?

Yes, I tested with my 32-inch monitor, it changes to small size in master.

Here is how to reproduce it:
- Change the font to some other say "Consolas"
- apply the new font
- Hit restore to default and apply settings.
- The new default font has a size less than 12 px, something like 8-9.

coder3101: image

PR #431, opened 2020-06-20
refactor(locale): remove locations in the translations

ouuan:
## Description

Remove locations in the translations in git, and if the author wants to see the source file at local, he/she can run `tools/updateTranslation.sh r` when editing and run `tools/updateTranslation.sh` after finished.

## Motivation and Context

The location changes are useless and big, which are not good for version control.

### DIFF
--- a/tools/updateTranslation.sh
+++ b/tools/updateTranslation.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+if [[ $1 == 'r' ]]; then
+	LOCATIONS=relative
+elif [[ $1 == 'a' ]]; then
+	LOCATIONS=absolute
+else
+	LOCATIONS=none
+fi
+
 declare -a SOURCES
 declare -a DIRS
 declare -a INCS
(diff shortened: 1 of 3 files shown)


coder3101:
What is use of location? What exactly does location in source points to? Just a general question.

coder3101:
If I have added a push button somewhere with some text, how do I make its russian and Chinese language, where to add it? I know `tr("Push me")`, what should I do after this. 

Since, I will not be involved in translation (no language known), I don't want to understand everything, just enough to guide others, probably new contributor. A simple explanation would be great.

ouuan:
> What is use of location? What exactly does location in source points to?

image

Used in the top-right window.

ouuan:
> If I have added a push button somewhere with some text, how do I make its russian and Chinese language, where to add it? I know `tr("Push me")`, what should I do after this.
> 
> Since, I will not be involved in translation (no language known), I don't want to understand everything, just enough to guide others, probably new contributor. A simple explanation would be great.

1. Compile first, so that the generated codes can be translated.
2. Run `tools/updateTranslation.sh`.
3. Open the `.ts` file in Qt Linguist and edit the translations.
4. Mark the translations as finished (<kbd>Ctrl+Enter</kbd>) if you think they are finished.

1 and 2 are not needed if they are done by others.

PR #438, opened 2020-06-21
feat & fix(auto-save): add auto save interval and fix #392

### LINKED ISSUE #392: False external file change when auto-save is enabled

ouuan:
## Description

1. Add auto-save time interval, which can be one of "After the last modification", "After the first modification" or "Without modification".
2. Fix #392.
3. Move the auto-save timer into the MainWindow, which is separated for each tab. This improves performance as it only saves one tab each time.

I think this is better than #412, not only for the functionalities but also for more clear logic and fewer codes.

## How Has This Been Tested?

On Arch Linux.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -13,11 +13,13 @@ With the portable version, you can easily store it in something like a USB disk,
 - Add the translation system and Simplified Chinese translation. (#377 and #384)
 - In-application UI style setting and built-in light/dark style. (#265 and #404)
 - Now testcases with empty outputs can also be checked. (#208 and #430)
+- Auto-save time interval.
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


coder3101:
You should also add the mention of "AutoSave types" Changelog

ouuan:
> You should also add the mention of "AutoSave types" Changelog

That's included in "Auto Save Interval".

PR #446, opened 2020-06-24
refactor(translation): make codes easier to maintain

ouuan:
##  Description

1. Use `TRKEY` in `PreferencesWindow` to avoid duplicate keys.
2. Convert inline functions in `PathItem` to member functions and use `tr` instead of `QCoreApplication::translate`.
3. Make `SettingsInfo` a class and use `Q_DECLARE_TR_FUNCTIONS` to use `tr` instead of `QCoreApplication::translate`.
4. Move `SettingsInfo.hpp` out of `generated` because it has nothing generated.
5. Remove an extra bracket in `EventLogger` which causes compilation error in platforms other than Linux, Windows and macOS, and causes a warning when updating the translations.

## Motivation and Context

1. Remove duplicates.
2. Use the shorter `tr` instead of the longer `QCoreApplication::translate`.
3. Make `SettingsInfo` more OOP.
4. Fix warnings.

## How Has This Been Tested?

On Arch Linux.

## Type of changes

- [x] Refactor (changes which affect the meaning of the code but neither fix a bug nor add a feature)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ ENDIF(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
 configure_file(cmake/version.hpp.in ${CMAKE_BINARY_DIR}/generated/version.hpp)
 configure_file(cmake/portable.hpp.in ${CMAKE_BINARY_DIR}/generated/portable.hpp)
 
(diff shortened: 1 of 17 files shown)


ouuan:
> Since this is a refactor change can you not make `revealInFileManager(QString path)` a free function inside Utils. We use it at 2/3 places.

That should be in another PR. Just like two features are not in a PR, two refactors are also not in a PR. This PR is for translation-related changes.

PR #448, opened 2020-06-24
feat(UpdateChecker): add proxy support

ouuan:
## Description

Now you can use proxy to check for updates.

## Motivation and Context

In some countries, it's easier to check for updates via proxy.

## How Has This Been Tested?

On Arch Linux, with clash and both Http and Socks5 proxy.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -16,6 +16,7 @@ With the portable version, you can easily store it in something like a USB disk,
 - Auto-save time interval.
 - Now you can auto-save the current session, besides saving it when the application exists. (#437 and #442)
 - Now the width of the input/output/expected of each test case is adjustable, and the maximum height of a test case can be set in the preferences. (#414 and #444)
+- Now you can use proxy to check for updates. (#448)
 
 ### Fixed
 
--- a/src/Settings/PreferencesPageTemplate.cpp
+++ b/src/Settings/PreferencesPageTemplate.cpp
@@ -74,26 +74,17 @@ PreferencesPageTemplate::PreferencesPageTemplate(QStringList opts, bool alignTop
         auto si = SettingsInfo::findSetting(name);
         if (si.depends.isEmpty())
             continue;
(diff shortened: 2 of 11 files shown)


StudyingFather:
Maybe we can add a option to disable proxy (direct connection).

That is, we have three options:

- disable proxy
- use system setting
- enable proxy

ouuan:
During the test on Windows, I found that we missed the OpenSSL DLLs in the installer, and the update checker always has the error "TLS initialization failed". I thought that's because of my internet connection before, till today I used the proxy and it still has this error.

ouuan:
> During the test on Windows, I found that we missed the OpenSSL DLLs in the installer, and the update checker always has the error "TLS initialization failed". I thought that's because of my internet connection before, till today I used the proxy and it still has this error.

It also happens on the AppImage. So maybe the users were never able to check for updates in the application.

neko-para:
> > During the test on Windows, I found that we missed the OpenSSL DLLs in the installer, and the update checker always has the error "TLS initialization failed". I thought that's because of my internet connection before, till today I used the proxy and it still has this error.
> 
> It also happens on the AppImage. So maybe the users were never able to check for updates in the application.

You can install openssl in Qt Installer, and the dlls are under $QTROOT/Tools/OpenSSL/$platform/bin

neko-para:
If the dlls are in the same directory of cpeditor, the error disappear.

ouuan:
> You can install openssl in Qt Installer, and the dlls are under $QTROOT/Tools/OpenSSL/$platform/bin

> If the dlls are in the same directory of cpeditor, the error disappear.

Please discuss OpenSSL related things in #449 and #451.

coder3101:
You should merge this PR after his approval.

neko-para:
Sorry, I'm outside yesterday. I'll test it right now.

neko-para:
I works fine on my machine. I test right proxy and wrong proxy and cpeditor works fine.

PR #451, opened 2020-06-25
feat(UpdateChecker): add a guide to install OpenSSL

### LINKED ISSUE #449: OpenSSL is not deployed, so it's not able to check for updates with the binary releases
**Describe the bug**

Can't check for updates with the binary releases.

**To Reproduce**

1. Download a binary release.
2. Run it.
3. Check for updates.
4. Get "TLS initialization failed".

**Screenshots**

image

**Environment**

Tested on both Arch Linux and Windows 10.

I believe we've never deployed OpenSSL, so the bug has been there since the update checker was added.

**Additional text**

On Windows, we can fix it by copying `Qt/Tools/OpenSSL/Win_x64/bin/*.dll` into the executable file directory.

ouuan:
## Description

Add a guide to install OpenSSL when the "TLS initialization failed" error happens in the update checker.

## Related Issue

This fixes #449.

## Motivation and Context

1. The user will be unable to check for updates in the application without OpenSSL.
2. It's hard and insecure to bundle the OpenSSL in the binary release, because OpenSSL is security-related, if we bundle it, it's hard to fix security problems. This is also why linuxdeployqt doesn't bundle OpenSSL in the AppImage (linuxdeploy/linuxdeploy-plugin-qt#57).

### DIFF
--- a/src/Telemetry/UpdateChecker.cpp
+++ b/src/Telemetry/UpdateChecker.cpp
@@ -103,7 +103,16 @@ void UpdateChecker::managerFinished(QNetworkReply *reply)
     if (reply->error())
     {
         LOG_ERR(INFO_OF(reply->errorString()));
-        progress->onUpdateFailed(reply->errorString());
+        auto error = reply->errorString();
+        if (error.contains("TLS initialization failed"))
+        {
+            error += "<br /><br />" +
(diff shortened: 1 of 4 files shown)


neko-para:
I think it's better to install ssl libraries

neko-para:
Especially on Windows

PR #455, opened 2020-06-28
feat(menu): add a link to issues in the main menu

ouuan:
## Description

Add a link to issues in the main menu for troubleshooting.

## Motivation and Context

Now many users are still writing comments in the CF blog instead of opening issues, hopefully this link in the menu can make more users open issues instead of asking elsewhere.

The user can also be redirected to FAQ/Manual/Telegram/Slack because of the new issue page:

image

### DIFF
--- a/src/appwindow.cpp
+++ b/src/appwindow.cpp
@@ -543,7 +543,7 @@ void AppWindow::on_actionManual_triggered()
         QUrl(tr("https://github.com/cpeditor/cpeditor/blob/%1/doc/MANUAL.md").arg(APP_VERSION), QUrl::TolerantMode));
 }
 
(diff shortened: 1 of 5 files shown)


ouuan:
> The URL opens "Known bugs", it doesn't troubleshoots them. So I think Troubleshooting is misleading.

If the user is not signed in, the new issue page is the login page.

So what's the change you suggest?

ouuan:
What's more, it's better to search the issues before opening a new issue.

coder3101:
"Report an Issue" or "Known Issues" is a better name for the option. From my windows experience Troubleshooting means, rectifying some issue with the application locally and presenting the fix actions or fix it automatically. Like Troubleshooting network connection.

Issue #413, opened 2020-06-18
Indent guide in editor


lakshits11:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
there should be indent guide lines so that we can see indentation properly like there in vs code.

**Describe alternatives you've considered**
none

**Additional context**

coder3101:
It will be better, if you could attach a screenshot of the line you get in VSCode. 

I guess you mean, a vertical line from starting `{` to its closing `}` for indentations.

If yes, its hard to implement. In QScintilla, this is called `IndentationGuides`. SEE `#3` [here](https://qscintilla.com/#simple_options/indentation)

lakshits11:
Yes, I mean that!!

sonulohani:
Hi @coder3101 

Can i work on this feature if nobody is working.

Thanks

sonulohani:
Hi @coder3101 Will https://github.com/cpeditor/cpeditor/pull/516 resolve this issue

coder3101:
> Hi @coder3101 Will #516 resolve this issue

No. We will just replace QCodeEditor with KSyntaxHighlighting based Code Editor. It will have just code folding and the ability to choose custom themes.

coder3101:
If you want, you can optimize the PR and your logic. Please note that the new code editor will be faster. The slow issue was attributed to some time taking code inside Editor line area, that has been resolved in both QCodeEditor and that PR.

sonulohani:
> If you want, you can optimize the PR and your logic. Please note that the new code editor will be faster. The slow issue was attributed to some time taking code inside Editor line area, that has been resolved in both QCodeEditor and that PR.

The above points has been taken into consideration. Please review the above PR.

PR #457, opened 2020-06-28
feat(IndentationGuide): Add indentation guide option in pref.

sonulohani:
Issue: cpeditor/cpeditor#413

image


sonulohani:
Build will fail because QCodeEditor doesn't have function available in the master

coder3101:
`Show Indentation Guides` will be a better name for the Settings.

coder3101:
You can also add customization like Indentation guide Style (Dashed, Dotted, Solid) and make sure that indent guides line colours are not hardcoded but used from style.xml (See Theme syles, [dracula.xml](https://github.com/cpeditor/cpeditor/blob/master/resources/styles/dracula.xml)) You can add a key/value and retrieve in QCodeEditor while painting.

PR #460, opened 2020-06-30
feat(submodules): update QtFindReplaceDialog and testlib

ouuan:
## Description

Now when you open the Find/Replace dialog with some text selected in the code editor, the text-to-find changes without clearing the undo history.


coder3101:
Changelog?

Issue #462, opened 2020-06-30
Allow user to choose the codec used by MessageLogger 


neko-para:
The MessageLogger now treats the program output as utf8, while some program may use the system locale(especially on windows).
We can detect the program's output or allow user to choose.
Or we can set the environment variable LANG and LANGUAGE before running program.
![image](https://user-images.githubusercontent.com/26452623/86102464-255c2500-baee-11ea-8cc4-e94d2b768254.png)

neko-para:
Closed via #469

PR #469, opened 2020-07-01
feat: Auto detect codec for compiler output

neko-para:
Add guessCodec to detect codec.

related issue: #462

## Type of changes

- [ ] New feature (changes which add functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/cpeditor/cpeditor/blob/master/CONTRIBUTING.md) document.
- [ ] I have tested these changes locally, and this fixes the bug/the new feature behaves as the expectation.
- [ ] The settings file in the old version can be used in the new version after this change.
- [ ] These changes only fix a single bug/introduces a single feature. (Otherwise, open multiple Pull Requests instead, unless these bugs/features are closely related.)
- [ ] The commit messages are clear and detailed. (Otherwise, use `git reset` and commit again, or use `git rebase -i` and `git commit --amend` to modify the commit messages.)
- [ ] These changes don't remove an existing feature. (Otherwise, add an option to disable this feature instead, unless it's necessary to remove this feature.)
- [ ] If there are changes of the text displayed in the UI, I have wrapped them in `tr()` or `QCoreApplication::translate()`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,6 @@ find_package(Qt5 COMPONENTS Widgets REQUIRED)
 find_package(Qt5 COMPONENTS Network REQUIRED)
 find_package(Qt5 COMPONENTS LinguistTools REQUIRED)
 find_package(Python3 COMPONENTS Interpreter REQUIRED)
-find_package(ICU COMPONENTS uc i18n REQUIRED)
 
 add_subdirectory(third_party/QCodeEditor)
 
@@ -211,8 +210,6 @@ target_link_libraries(cpeditor PRIVATE LSPClient)
 target_link_libraries(cpeditor PRIVATE QCodeEditor)
 target_link_libraries(cpeditor PRIVATE Qt5::Network)
 target_link_libraries(cpeditor PRIVATE Qt5::Widgets)
-target_link_libraries(cpeditor PRIVATE ICU::uc)
-target_link_libraries(cpeditor PRIVATE ICU::i18n)
 target_link_libraries(cpeditor PRIVATE QtFindReplaceDialog)
 target_link_libraries(cpeditor PRIVATE SingleApplication)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(Qt5 COMPONENTS Widgets REQUIRED)
 find_package(Qt5 COMPONENTS Network REQUIRED)
 find_package(Qt5 COMPONENTS LinguistTools REQUIRED)
 find_package(Python3 COMPONENTS Interpreter REQUIRED)
+find_package(ICU COMPONENTS uc i18n REQUIRED)
 
 add_subdirectory(third_party/QCodeEditor)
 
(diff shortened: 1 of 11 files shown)


ouuan:
I replaced the hardcoded codec list with `QTextCodec::availableMibs()` and tested, then the first available codec (US-ASCII) is used, there are lots of invalid characters in the result, but `state.invalidChars` is 0 :thinking: (There are Chinese in the compiler messages, and it displays fine in UTF-8.)

ouuan:
BTW, if I change the hardcoded list to `{"KOI8-RU", "UTF-8", "GB2312"}`, it crashes because `KOI8-RU` is not available on my system.

ouuan:
If I change the hardcoded list to `{"GB2312", "UTF-8", "KOI8-RU"}`, it uses `GB2312` and displays invalid characters.

neko-para:
It seems that libicu provides encoding detection. I'll try using it.

ouuan:
Can we let the user choose the codec?

neko-para:
> Can we let the user choose the codec?

Then we have to add three option, each language use one. Different compilers may use different codec.

neko-para:
@ouuan Can you help me test the detection? It's really hard for me to make icu version the same as qt used. But it may be easy for arch user. On my ubuntus, the program randomly crushes. I print the error status of icu functions and it tells me that the error code is bogus. I think it's because the wrong version of icu.

ouuan:
I don't think it's a good idea to introduce a new library for this small function. I'd rather add codec options in the settings.

neko-para:
After all we can use this to guess the codec of source.

ouuan:
It works on my Arch Linux.

Logs:

```log
[2020-07-01T21:11:37.050][INFO ][          guessCodec          ][      Util.cpp      ](44)::Find 7 matches
[2020-07-01T21:11:37.050][INFO ][          guessCodec          ][      Util.cpp      ](48)::Highest confidence is 100
[2020-07-01T21:11:37.050][INFO ][          guessCodec          ][      Util.cpp      ](57)::Pick UTF-8
```

My icu version is 67.1.

neko-para:
Actually Qt use libicu, but I don't know how to use it.

neko-para:
On Linux the libicuxxx.so are installed together with qt, but no headers.

ouuan:
I don't think it's good to assume the locale of the compiler and the application locale are the same, and Chinese codec is not necessarily GB18030 :thinking:

neko-para:
So use `LANG=C`? I don't know any more solutions.

ouuan:
Why not let the user set the codec?

neko-para:
Then add a new item called `CodecItem`? Sounds ok. I'll try it.

ouuan:
> Then add a new item called `CodecItem`?

I think combo box works.

ouuan:
`param: "QTextCodec::availableCodecs()"`

neko-para:
`QTextCodec::availableCodecs()` return too many results. I don't think it's a good choice.

ouuan:
`"param": "([] { QStringList names; for (auto mib : QTextCodec::availableMibs()) { names.push_back(QString::fromLocal8Bit(QTextCodec::codecForMib(mib)->name())); } names.sort(Qt::CaseInsensitive); names.removeDuplicates(); return names; })()"`

This seems much better.

ouuan:
> Then add a new item called `CodecItem`?

However, we can still make a `CodecWrapper` inherited from `ComboBoxWrapper` so that it's reusable.

ouuan:
I think it's better to put Compiler Output Codec at the bottom of that page :thinking: It's useful for only a few people, and it's a QComboBox while others are QLineEdits.

neko-para:
Sounds reasonable.

PR #470, opened 2020-07-01
ci: fix empty displayed version in GitHub Actions

ouuan:
## Description

Fix empty display version in GitHub Actions caused by no commit history.


ouuan:
> Depth zero will also speed clone speed as less bytes need to be downloaded.

Depth zero means all histories so that this bug can be fixed, which slows down the cloning process but it's can almost be ignored.

PR #471, opened 2020-07-01
build(aur): use Ninja to build

ouuan:
## Description

Use Ninja to build in the PKBUILD for AUR.

## Motivation and Context

1. Ninja doesn't spam in the terminal like `make`.
2. It is a little bit faster.

## How Has This Been Tested?

Used `makepkg -si` to install the PKGBUILD.

### DIFF
--- a/.ci/aur/git/.SRCINFO
+++ b/.ci/aur/git/.SRCINFO
@@ -7,6 +7,7 @@ pkgbase = cpeditor-git
 	license = GPL3
 	makedepends = cmake
 	makedepends = git
+	makedepends = ninja
 	makedepends = python3
 	makedepends = qt5-tools
 	depends = qt5-base>=5.15.0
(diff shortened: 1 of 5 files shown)


ouuan:
I'm wondering whether always color by default could cause some problems, like when building in the IDE :thinking: However, I always build in the terminal so I can't test.

ouuan:
Maybe it's a better choice to disable always color by default and enable it in the PKBUILD.

ouuan:
But if always color wouldn't cause problem unless you redirect the output to a file, it's good to enable it by default so that people using Ninja don't have to enable it manually.

coder3101:
I don't think color is that big of a deal. Let's not worry about Build steps. We should keep it default as offered by CMAKE.

ouuan:
> I don't think color is that big of a deal.

If it shows color in unsupported place, the output will be in a mess.

> Let's not worry about Build steps.
> We should keep it default as offered by CMAKE.

These are somehow conflicted, because the default now is not the default offered by CMake. What do you mean by "not worry"?

Issue #269, opened 2020-04-29
More settings


ouuan:
![near-future plan](https://user-images.githubusercontent.com/30581822/77845345-c500d300-71e0-11ea-99c0-a449c48d88ff.png)

_Originally posted by @ouuan in https://github.com/cpeditor/cpeditor/issues/166#issuecomment-605607699_

Some settings of this plan are neither added nor requested in an issue, so I open this for them.

neko-para:
Why this issue is still open? It seems that all the settings have been added.

ouuan:
> It seems that all the settings have been added.

No.

ouuan:
Now there are the following places to choose a file path (or directory):

1. Open Files
2. Save File / Save As
3. Open Contest
4. Export/Import settings
5. Export/Load session
6. Export/Load snippets
7. Choose C++ template
8. Choose Java template
9. Choose Python template
10. Choose various executable files in the settings, including CF Tool and Clang Format.
11. Load a single test case from file
12. Add pairs of test cases from files
13. Choose a custom checker

My plan is to give the user 13 (or more) default paths, and the setting for each action consists of two options, the default path to use and the default path to change. The default path to use is used as the default path in the QFileDialog, and the default path to change is changed after choosing the path. Both of them can be omitted, i.e. the default path in the QFileDialog can be empty and there can be nothing changed after choosing the path. Each default path can also be changed manually in the settings. In addition to this, the default path for saving a file can be overridden by "Default File Paths For Problem URLs".

This is a little bit too complicated, however, the user can ignore it and use the default settings. Are there any better ideas?

ouuan:
Now except for the log-file-related settings, other unimplemented settings have their own issues. Because the log file is very basic, now I think it's better not to have those settings.

So all things are finished.

PR #483, opened 2020-07-06
feat(Settings): add a flexible system to set and change default paths

ouuan:
## Description

Add a system to set and change default paths for all choosing path actions.

You can read the tooltips and try it to understand how it works.

It doesn't apply to PathItem because the default path of a PathItem is the old value of it.

## Related Issue

It's a part of #269.

## Motivation and Context

Before this, only saving/opening a file can remember the last chosen path, and the rule is hard-coded —— the default path is always the last path you choose.

Now it applies to nearly all choosing path actions, and the rule is very flexible. You can use either a manually set path or an automatically changed path. You can use the same default path for several actions, or even change another action's default path when choosing a path.

## How Has This Been Tested?

On Arch Linux.

image

## Type of changes

## Checklist

### DIFF
--- a/src/Settings/DefaultPathManager.cpp
+++ b/src/Settings/DefaultPathManager.cpp
@@ -41,7 +41,7 @@ void DefaultPathManager::changeDefaultPathForAction(const QString &action, const
 
     if (!SettingsManager::contains(settingsKey, true))
     {
-        qDebug() << "Unkown Action:" << action;
+        qDebug() << "Unknown Action:" << action;
         LOG_ERR("Unknown Action");
         return;
     }
--- a/src/Settings/DefaultPathManager.cpp
+++ b/src/Settings/DefaultPathManager.cpp
@@ -1,0 +1,155 @@
+/*
+ * Copyright (C) 2019-2020 Ashar Khan <ashar786khan@gmail.com>
+ *
+ * This file is part of CP Editor.
+ *
+ * CP Editor is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * I will not be responsible if CP Editor behaves in unexpected way and
+ * causes your ratings to go down and or lose any important contest.
+ *
+ * Believe Software is "Software" and it isn't immune to bugs.
+ *
+ */
+
+#include "Settings/DefaultPathManager.hpp"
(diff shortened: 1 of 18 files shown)


coder3101:
Do you think it is important to add article "The" to all titles?

coder3101:
It also very redundant to have full title like "The default path to something" instead we should have "Something" because the title of page itself says default path.

ouuan:
Do you mean 4b69b99?

coder3101:
> Do you mean 4b69b99?

Yes; I only checked Screenshot and missed the commit.

src/Settings/DefaultPathManager.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+void DefaultPathManager::changeDefaultPathForAction(const QString &action, const QString &path)
+{
+    LOG_INFO(INFO_OF(action) << INFO_OF(path));
+
+    const auto settingsKey = QString("Default Path/Action/%1/Changes").arg(action);
+
+    if (!SettingsManager::contains(settingsKey, true))
+    {
+        qDebug() << "Unkown Action:" << action;
```
neko-para: Maybe remove this?

ouuan: Why? It's just like unknown settings. All actions are strings, it's very possible to have a typo. It's more obvious to show this in stderr.

neko-para: Ok

neko-para:
It seems good. Maybe I could add some function to make querying, modifying untraced settings easier.

ouuan:
> Maybe I could add some function to make querying, modifying untraced settings easier.

Do you have any plans? I think it's easy enough now :thinking:

neko-para:
I mean sth. like `SettingsHelper::dir("Default Path").dir("Action").dir(xxx).str("Uses")`

ouuan:
> I mean sth. like `SettingsHelper::dir("Default Path").dir("Action").dir(xxx).str("Uses")`

Do you think that's easier?

neko-para:
Actually I wonder if the default path could use Object instead of VariantList

ouuan:
> Actually I wonder if the default path could use Object instead of VariantList

Object is better, but we need a new widget for it.

neko-para:
Yes. Object allow recursive. It will be a big work

neko-para:
Maybe we can use Property system in QObject?

PR #488, opened 2020-07-07
feat(Settings): add an option to format on auto-save

ouuan:
## Description

Add an option to format on auto-save and move auto-format settings to Clang Format.

## Related Issue

Part of #269.

## Motivation and Context

It could be either useful or annoying to format on auto-save, so we can let the user choose it.

## How Has This Been Tested?

Tested on Arch Linux with four possible settings combination.


src/Settings/settings.json
```diff
@@ -59,6 +59,18 @@
         "ui": "QPlainTextEdit",
         "tip": "The Clang Format style options, which are often saved in a .clang-format configuration file.\nYou can learn about it at <https://clang.llvm.org/docs/ClangFormatStyleOptions.html>."
     },
+    {
+        "name": "Clang Format/Format On Manually Save",
```
coder3101, line 63: *Format on Manual Save*

ouuan: I'm wondering what's the grammar here, because in Oxford Advanced Learner's Dictionary, "save" in noun form has only one meaning: "an action by the goalkeeper that stops a goal being scored". So I guess this is an informal use, where "save" is actually a verb. As a reference, many DOM events are not "on<noun>", including "onabort", "onkeydown", "onselect", "onsubmit" and many more.

ouuan: However, maybe it should be "save manually" instead of "manually save" :thinking: Or "saving manually", if you prefer correct grammar. I don't think "manual save" is correct, since "save" as a noun doesn't mean saving a file.

ouuan: Since both are incorrect grammar, maybe treating "save" as a noun sounds better :thinking:

PR #491, opened 2020-07-09
feat: elide test cases when they are too long

### LINKED ISSUE #487: Only display first <limit> characters if test cases are too long, instead of not loading them

[The testcase file [name.in] || (pasted content) contains more than 5000000 characters, so it only display first 500000 char. The full input is available at name.in. You can change the length limit in Preferences->Advanced->Limits->Test Case Length Limit]

[The output contains more than 500000 characters, so it only display first 5000000. The full output is available at name.out. You can set the output length limit in Preferences->Advanced->Limits->Output Length Limit]

ouuan:
## Description

Now the test cases loaded from files and program output are elided and read-only if they are too long.

## Related Issue

This closes #487.

## Motivation and Context

Users may want to test on big test cases while the performance is good.

## How Has This Been Tested?

On Arch Linux.

image

## Type of changes

## Checklist

### DIFF
--- a/src/Widgets/TestCaseEdit.cpp
+++ b/src/Widgets/TestCaseEdit.cpp
@@ -92,8 +92,8 @@ void TestCaseEdit::modifyText(const QString &text, bool keepHistory)
 {
     this->text = text;
 
-    const int limit = role == Output ? SettingsHelper::getOutputDisplayLengthLimit()
-                                     : SettingsHelper::getLoadTestCaseFileLengthLimit();
+    const int limit =
+        role == Output ? SettingsHelper::getOutputDisplayLengthLimit() : SettingsHelper::getLoadTestCaseLengthLimit();
 
     QString displayText;
 
@@ -110,9 +110,8 @@ void TestCaseEdit::modifyText(const QString &text, bool keepHistory)
         setReadOnly(true);
 
         const QString name = role == Input ? tr("input") : (role == Output ? tr("output") : tr("expected"));
-        const QString setLimitPlace = role == Output
-                                          ? tr("Preferences->Advanced->Limits->Output Display Length Limit")
-                                          : tr("Preferences->Advanced->Limits->Load Test Case File Length Limit");
+        const QString setLimitPlace = role == Output ? tr("Preferences->Advanced->Limits->Output Display Length Limit")
+                                                     : tr("Preferences->Advanced->Limits->Load Test Case Length Limit");
 
         displayText = tr("The %1 is too long, only the first %2 characters are shown. You can set the length limit in "
                          "%3.\n\n%4...")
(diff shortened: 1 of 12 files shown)


src/Widgets/TestCaseEdit.cpp
```diff
@@ -65,19 +80,64 @@ void TestCaseEdit::dropEvent(QDropEvent *event)
...
+    {
+        LOG_INFO("Too long: " << INFO_OF(role) << INFO_OF(text.length()));
+
+        setReadOnly(true);
+
+        const QString name = role == Input ? tr("input") : (role == Output ? tr("output") : tr("expected"));
+        const QString setLimitPlace = role == Output ? tr("Preferences->Advanced->Limits->Output Display Length Limit")
+                                                     : tr("Preferences->Advanced->Limits->Load Test Case Length Limit");
+
+        displayText = tr("The %1 is too long, only the first %2 characters are shown. You can set the length limit in "
```
coder3101, line 116: Why do you think it should be prepend  instead of append?

ouuan: Because I want to let the user know this happens. Otherwise it's very possible that they don't know why the input is read-only.

ouuan: Or we can move this to the message logger.

coder3101: > Or we can move this to the message logger.

Yes, that will be better because showing long texts in Input/Output is not good ideas as user needs to scroll to read it.

ouuan: > user needs to scroll to read it

I think the user can simply ignore it and only have to read it once, but if it's in the mssage logger, the user needs to scroll to read other messages.

coder3101: Why not set shorten the message to something like: 
"Showing only the first 200 characters"
<200 characters of stdout>

In message logger, you can say in a warning:
"[Limits Warnings] : stdout exceeded maximum output limit set in Preferences->Advanced->Limits->Output Limit"

ouuan: Do you mean splitting the message into two parts, one in the TestCaseEdit and another in the message logger?

coder3101: Yes

ouuan: Maybe we can move this to the message logger, and make the "where to set limit" messages optional in the settings.

coder3101: I don't think a new settings for this is a good idea, This way our settings will be bloated.

ouuan: The HTML title seems like a good solution.

PR #500, opened 2020-07-13
feat: Only update generated file when needed

neko-para:
Now genSettings.py will read the old generated file and won't update
them if there aren't any changes.


ouuan:
build -> delete `SettingsHelper.hpp` and `SettingsInfo.cpp` -> build -> compilation error

Is it very frequent that `settings.json`/`genSettings.py` are changed but the generated files are not?

ouuan:
It doesn't work if `SettingsHelper.hpp` and `SettingsInfo.cpp` is not there.

neko-para:
It seems that quite hard to solve it.

PR #502, opened 2020-07-14
ci: fix release checksums

ouuan:
## Description

Make `dn()` in multi-lines.

## Motivation and Context

Bash functions need to be in multi-lines.

image


ouuan:
> So wasn't it working prior to this commit?

There's no new version after #411, so it has been never executed before.

PR #527, opened 2020-07-26
feat: open recent files

ouuan:
## Description

Add "Open Recent Files" in the menu.

## How Has This Been Tested?

On Arch Linux.

## Type of changes

## Checklist

@IZOBRETATEL777 Please add Russian translations.

### DIFF
--- a/src/appwindow.cpp
+++ b/src/appwindow.cpp
@@ -74,6 +74,8 @@ AppWindow::AppWindow(bool noHotExit, QWidget *parent) : QMainWindow(parent), ui(
                 openRecentFilesMenu->addAction(recentFile, [this, recentFile] { openTab(recentFile); });
             }
         }
+        openRecentFilesMenu->addSeparator();
+        openRecentFilesMenu->addAction(tr("Clear Recent Files"), [] { SettingsHelper::setRecentFiles({}); });
     });
 
     Core::StyleManager::setDefault();
--- a/src/appwindow.cpp
+++ b/src/appwindow.cpp
@@ -58,6 +58,24 @@ AppWindow::AppWindow(bool noHotExit, QWidget *parent) : QMainWindow(parent), ui(
     allocate();
     setConnections();
 
+    auto separator = ui->menuFile->insertSeparator(ui->actionSave); // used to insert openRecentFilesMenu
+    auto openRecentFilesMenu = new QMenu(tr("Open Recent Files"), ui->menuFile);
+    ui->menuFile->insertMenu(separator, openRecentFilesMenu);
+    connect(openRecentFilesMenu, &QMenu::aboutToShow, [this, openRecentFilesMenu] {
+        openRecentFilesMenu->clear();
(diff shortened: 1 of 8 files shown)


src/appwindow.cpp
```diff
@@ -58,6 +58,24 @@ AppWindow::AppWindow(bool noHotExit, QWidget *parent) : QMainWindow(parent), ui(
...
     setConnections();
 
+    auto separator = ui->menuFile->insertSeparator(ui->actionSave); // used to insert openRecentFilesMenu
+    auto openRecentFilesMenu = new QMenu(tr("Open Recent Files"), ui->menuFile);
+    ui->menuFile->insertMenu(separator, openRecentFilesMenu);
+    connect(openRecentFilesMenu, &QMenu::aboutToShow, [this, openRecentFilesMenu] {
+        openRecentFilesMenu->clear();
+        if (SettingsHelper::getRecentFiles().isEmpty())
+        {
+            openRecentFilesMenu->addAction(tr("No file is opened recently"))->setDisabled(true);
```
coder3101, line 68:
```suggestion
            openRecentFilesMenu->addAction(tr("No Recent Files"))->setDisabled(true);
```

It seems more short and straightforward.

ouuan: I tried to explain what "recent file" is.

coder3101: Isn't word **recent files** self explanatory?

ouuan: It's not necessary to explain "recent files", but here, I think we can explain it via a warning. Though it's not necessary, it's not completely meaningless. "Recent files" doesn't point out what is "recent". Recently edited? Recently opened? Recently used elsewhere instead of in CP Editor?

ouuan: What's more, "No recent files" is a noun phrase instead of a sentence, which sounds weird to me. I think "There is no recent file" is much better, but "No file is opened recently" is even better.

coder3101: You can see what VS Code or other tools use for this. "Recent files" in general in context of editor means, recently opened files in that editor.

ouuan: > You can see what VS Code or other tools use for this.

I don't know how to completely clear the recent files in VS Code. In Sublime Text, there's no this sentence (i.e. hint for no file is opened recently). In Qt Creator, the "Open Recent Files" menu is disabled when no file is opened recently.

I still don't know what's wrong with "No file is opened recently". I think that's more clear than "No recent files".

coder3101: Nothing is wrong, it's just too long. BTW, can't we also make "Open Recent files" disabled, if no recent file is found?

ouuan: I think it's more user-friendly to display a message for the reason that Open Recent Files is not available.

coder3101: Okay, it's resolved then.

PR #539, opened 2020-07-30
feat: Use contest dialog to open contest

### LINKED ISSUE #467: QWizard for Opening Contest
A new contest is created using "Open Contest" menu option and it launches the Directory picker followed by a Number Input dialog followed by a QCombo box for language selection. This sequencial flow can be better managed and be made verbose with a [QWizard](https://doc.qt.io/qt-5/qwizard.html)

Open a QWizard for choosing all Contest-related parameters.

Use plain old style

**Additional context**

image

See a small and ugly number picker.

coder3101:
Use contest dialogs to open Contests

## Description
This commit adds a dialog widget that can be used to open contest. 

## Related Issue
Fixes #467 

## Motivation and Context
This is better UI than last one.

## How Has This Been Tested?
Win 10 

image

## Type of changes

## Checklist

- [x] If there are changes in the text displayed in the UI, I have wrapped them in `tr()` or `QCoreApplication::translate()`.

**Required translations** @IZOBRETATEL777 and @sadykhzadeh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,4 +57,8 @@ CMakeSettings.json
 # Dependency directories
 node_modules/
 
+# Compile commands for LSP
+compile_commands.json
+
+# Translations
 resources/translations/*.qm
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,8 @@ add_executable(cpeditor
     src/Util/Util.cpp
     src/Util/Util.hpp
 
+    src/Widgets/ContestDialog.cpp
+    src/Widgets/ContestDialog.hpp
     src/Widgets/DiffViewer.cpp
     src/Widgets/DiffViewer.hpp
     src/Widgets/TestCase.cpp
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ It makes your competitive coding life easier :grin: by automating many things fo
 
 ## Get help
 
-1. Read the [manual](doc/MANUAL.md) and the [FAQ](#faq) first, also go through the menus and preferencecs to see if there are what you are looking for.
+1. Read the [manual](doc/MANUAL.md) and the [FAQ](#faq) first, also go through the menus and preferences to see if there are what you are looking for.
 2. Ask Google first if it's not very editor-related. (e.g. How to install Clang Format? What does this compilation error mean?)
 3. Search in the [Issues](https://github.com/cpeditor/cpeditor/issues) and make sure your bug report/feature request is not duplicated.
 4. Please follow the issue template and provide detailed information, it will help us to give you better feedback.
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Added
 
 - Open recent files. (#527)
+- Use Dialog box to open new contests (#539 & #467)
 
 ### Fixed
 
--- a/src/Settings/DefaultPathManager.cpp
+++ b/src/Settings/DefaultPathManager.cpp
@@ -33,7 +33,7 @@ QString DefaultPathManager::defaultPathForAction(const QString &action)
     return result;
 }
 
-void DefaultPathManager::changeDefaultPathForAction(const QString &action, const QString &path)
+void DefaultPathManager::setDefaultPathForAction(const QString &action, const QString &path)
 {
     LOG_INFO(INFO_OF(action) << INFO_OF(path));
 
@@ -67,7 +67,7 @@ QString DefaultPathManager::getExistingDirectory(const QString &action, QWidget 
 {
     const auto result = QFileDialog::getExistingDirectory(parent, caption, defaultPathForAction(action), options);
     if (!result.isEmpty())
-        changeDefaultPathForAction(action, result);
+        setDefaultPathForAction(action, result);
     return result;
 }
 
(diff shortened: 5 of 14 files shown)


coder3101:
> 
> 
> I'm not sure what the contest name is used for, but if it's for the new directory name:
> 
You are correct, it used to create new directory. Like my root for CP is say `/home/ashar/codeforces`, that remains. I can create a directory by setting a contest name as `1145`. So all my sample codes will be in `/home/ashar/codeforces/1145/`

>     1. It's not easier to create a new directory in CP Editor instead of in the file dialog;

I don't get it. 

>     2. It's possible that the user has already created the directory.
 Yes, possible. It will not create a directory in that case.

coder3101:
> 1.  I think it's better to select the directory of the contest instead of its parent. When I use this function, I usually have that directory created. Because, if I'm participating in an online contest, I'll parse it by Competitive Companion, and save it by "default path for problem URL"; when I use "Open Contest", it's usually an offline contest, I have already downloaded the statements and examples at local. The only benefit for separate root and contest name is that the root can be saved instead of the contest directory, but it should be easy to navigate to the parent directory in the file dialog.

Okay, a better solution will be to have a check box in the dialog that says something like "Use Contest root only" (if clicked the contest name be inactive/disabled). We can add a useful tooltip for the checkbox.


>     I mean, you can easily create a new directory in the file dialog.

Yes we can, but it's easier other way. The last contest root is preserved and when creating contest you can only use `tabs` and keyboard strokes to quickly create contest directory. Never having to open a file picker dialog or other things.


>     But the user needs to re-enter the contest name.

See response above

> 
> 2.  You don't have to write the destructor. Qt does the destruction, if you have either set the parent of the widgets or added the widgets/layouts to another layout. The same for AppWindow and MainWindow, but they were there at the very beginning, so I was lazy to change them.

Okay, I was aware but didn't cross checked because non of Qt examples has destructors. I'll remove them in my next push.

Maybe we can open an issue for new contributors for this refactoring.

> 
> 
> 3.  You should use the default path manager instead of a new setting.
> 

Okay I'll do it. It would be nice if you can show me where you have used so I can use it in similar way.

> 4.  Though Qt has Implicit Sharing, it's more clear to use `const &` whenever possible. Not for performance, but it makes codes easier to understand if we don't change the arguments, and `const &` enforces this. It's also hard to remember where `const &` is not needed, so it's better to use it whenever possible.
> 

I was about to do same, `const&` has many other benefits too.

> And a suggestion: Recently, I realized that we can declare the elements which are only used in the constructor (for example, most layouts and some widgets) in the constructor instead of as class members. That can make the codes much shorter.

Do you mean we create say `auto* layout = new QHBoxLayout(this)` in constructor. If this is what you mean, I think because we don't need to free stuff, this is safe and better unless of course layout is required in other member functions.

I will change this, again an issue for refactoring by new contributor for this can be opened.

ouuan:
> a better solution will be to have a check box in the dialog that says something like "Use Contest root only" (if clicked the contest name be inactive/disabled).

OK. But I prefer "Use Subdirectory"/"Use a subdirectory of the root directory to save the contest" and "Subdirectory Path" ("Path" instead of "Name" because it can be something like `CF/1172`), instead of "Use contest root only" and "Contest Name".

> It would be nice if you can show me where you have used so I can use it in similar way.

https://github.com/cpeditor/cpeditor/pull/539/files#diff-c4c2b0d89b42dc0584939e7dcc3efec5L656

https://github.com/cpeditor/cpeditor/blob/d56246714142c3fc2c8b5b1561a8fd1e85a48ff6/tools/genSettings.py#L106-L116

ouuan:
> > a better solution will be to have a check box in the dialog that says something like "Use Contest root only" (if clicked the contest name be inactive/disabled).
> 
> OK. But I prefer "Use Subdirectory"/"Use a subdirectory of the root directory to save the contest" and "Subdirectory Path" ("Path" instead of "Name" because it can be something like `CF/1172`), instead of "Use contest root only" and "Contest Name".

Or, the user can simply leave the "Subdirectory Path" empty if they don't want to use it. Maybe this hint can be in the tooltip.

coder3101:
>     3\. You should use the default path manager instead of a new setting.

I have added that as a **cached variable** rather than a default path? It is better being in settings. It is like setting we use to check if CP Editor was maximized when exiting. So next time we make it open in that mode.

ouuan:
> I have added that as a **cached variable** rather than a default path? It is better being in settings. It is like setting we use to check if CP Editor was maximized when exiting. So next time we make it open in that mode.

I'm not sure why you want to use a new setting instead of the default path manager. If you want to use it without the file dialog, you can use `QString DefaultPathManager::defaultPathForAction(const QString &action)`.

coder3101:
> QLayout: Attempting to add QLayout "" to Widgets::ContestDialog "", which already has a layout

What could be source for this warning?

coder3101:
If you want you can make it use `defaultPathManager`, I will do it some other day.

coder3101:
> If you want you can make it use `defaultPathManager`, I will do it some other day. (or you can quickly do it and merge the PR)

ouuan:
> What could be source for this warning?

https://github.com/cpeditor/cpeditor/issues/541#issuecomment-667481469

ouuan:
I'll directly push some commits instead of giving suggestions.

coder3101:
@ouuan Please check https://github.com/cpeditor/cpeditor/runs/943185588

ouuan:
> @ouuan Please check https://github.com/cpeditor/cpeditor/runs/943185588

I've opened 2m/arch-pkgbuild-builder#7 to fix it. We can either use `ouuan/arch-pkgbuild-builder@master` or wait until that PR is merged.

PR #547, opened 2020-08-04
ci: include .git in full source

ouuan:
## Description

Include `.git` in the full source.

Because the full source is of a release, the commit history is not needed, so we can fetch with depth 1.

## Motivation and Context

If you build CP Editor without any commit, `GIT_COMMIT_HASH` will be empty.

## How Has This Been Tested?

In my fork.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
-          fetch-depth: 0
 
       - name: Install Python 3.7
         uses: actions/setup-python@v1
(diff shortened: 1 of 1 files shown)


ouuan:
> Git history will grow and in near 1-2 years its size will be so big that it's most git stuff than code stuff.
> 
> I recommend one of two:
> 
>     * Manually clone submodules with zero depth and cpeditor with some threshold depth say 50.
> 
>     * Add a VERSION.txt inside source code and let cmake configure version from their if a file with that name exist. We can generate and include that text file from CI.

Depth 0 is not a parameter of git, it's a parameter of `actions/checkout`, which means no depth limit (i.e. all history).

The default value of the `delpth` parameter of `actions/checkout` is `1`.

In short, now there's no history except for the latest commit.

PR #554, opened 2020-08-08
feat(Detached Run): Choose terminal for execution on Linux

### LINKED ISSUE #474: Allow user to choose terminal for Detached Execution
**Describe the bug**

On pressing the specified key binding for run detached mode, i.e ctrl+alt+D, the editor minimises instead of opening the terminal. I think this should be fixed.

**Enhancement**

For ubuntu users, the xterm doesn't come pre-installed. We get the genome-terminal. So a support for that could also be added instead of the only available xterm. One would naturally use the terminal of his choice to have the its saved configured features.

**Environment**

- App version: 6.4.7
- Git commit hash: e352209
- Build time: Jun 19 2020 05:17:51
- OS: Linux

coder3101:
## Description
Add inside settings->Actions->Detached Execution an option to choose the terminal for detached execution.

## Related Issue
Fixes #474 

## Motivation and Context
xterm isn't present on every linux. If it is present, its  UI us very old and it isn;t widely used console.

## How Has This Been Tested?
Arch Linux

## Type of changes

- [x] New feature

## Checklist


coder3101:
Tested and found to be working fine on:

- XTerm
- Konsole

`gnome-terminal` has an issue where it launches 2 terminal instances, in one of them it runs program and other is simply bash prompt.

src/Core/Runner.cpp
```diff
@@ -116,24 +116,28 @@ void Runner::runDetached(const QString &tmpFilePath, const QString &sourceFilePa
...
     }
-    runProcess->setProgram("xterm");
-    runProcess->setArguments({"-e", getCommand(tmpFilePath, sourceFilePath, lang, runCommand, args) +
-                                        "; read -n 1 -s -r -p '\nExecution Done\nPress any key to exit'"});
-    LOG_INFO("Xterm args " << runProcess->arguments().join(" "));
+    runProcess->setProgram(terminal);
+    auto quotedCommand = getCommand(tmpFilePath, sourceFilePath, lang, runCommand, args);
+    QStringList execArgs = {
+        "-e", "/bin/bash -c '" + quotedCommand +
+                  " ; echo \"\nProgram completed with exit code $?\nPress any key to exit\" ; read -n 1'"};
```
ouuan, line 137:
```suggestion
    QStringList execArgs = {"-e", QStringLiteral("/bin/bash -c '%1 ; echo \"\n%2\" ; read -n 1'")
                                      .arg(quotedCommand)
                                      .arg(tr("Program finished with exit code %1\nPress any key to exit").arg("$?"))};
```

`.arg("$?")` is used to prevent translators from translating `$?`.

ouuan: I'm not very sure whether UTF-8 is always supported in the terminal :thinking: It works fine on my Arch Linux with the Chinese translation of this.

coder3101: Most terminal emulators on Linux do support them. It also depends if fonts are installed. But I guess most people using Chinese or Russian as their language will have UTF-8 support in their emulator.

coder3101:
> I thought that the command for different terminal emulators could be very different, but it turns out that they are nearly the same. So instead of providing three choices, we can let the user set `xterm -e`, `gnome-terminal --`, etc. in a QLineEdit. It enables users to use more terminal emulators and solves this problem: `# Option “-e” is deprecated and might be removed in a later version of gnome-terminal. # Use “-- ” to terminate the options and put the command line to execute after it.`.

I agree but then our Documentation is incomplete and adding something takes user input to change terminal is a bad idea for newbie users. 

Maybe we can add a checkbox that says "Use custom terminal emulator", which when clicked makes `QLineEdit` enabled for editing and Combobox disabled.  If unchecked the Combobox will be visible and user can change terminals just by choosing one from the list. It's like preset.

Gnome-terminal deprecating `-e` can be handled by adding a conditional in code.


> And we can remove `xterm` in the optdepends of the AUR package.

Yes, I will change that in my next commit.

ouuan:
> I agree but then our Documentation is incomplete and adding something takes user input to change terminal is a bad idea for newbie users.

It can be in the tooltips. And the documentation for preferences is on its way in cpeditor/cpeditor.github.io#53.

coder3101:
Okay, I'll change that.

src/Core/Runner.cpp
```diff
@@ -116,24 +116,28 @@ void Runner::runDetached(const QString &tmpFilePath, const QString &sourceFilePa
...
+
+    LOG_INFO("Using: " << terminal << " on Linux");
     QProcess testProcess;
-    testProcess.start("xterm", {"-v"});
+    testProcess.start(terminal, {"-v"});
     bool finished = testProcess.waitForFinished(2000);
     if (!finished || testProcess.exitCode() != 0)
     {
-        emit failedToStartRun(runnerIndex, tr("Please install xterm in order to use Detached Run."));
+        emit failedToStartRun(runnerIndex, QString(tr("Please install %1 in order to use detached run or change "
```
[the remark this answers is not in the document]
coder3101: Done

coder3101:
I think setting the ability to choose the terminal command is way too much complicated* for the user. Maybe we can just add some more presets like for `xfce-terminal` and `deepin-terminal`. For now its fine, In future we can provide more flexibility on this part.

[1] complicated* 
Complicated because a typical detached execution is processed like this. CP Editor forks a Terminal, which in turn forks a shell (BASH) in this case but can be different if user sets those commands. Shell executes our program and then echos and waits for `read -n 1`, as pointed out earlier the `read` command depends on the shell `-n` is not available in all shells except `bash`. Then comes different syntax for maybe different terminal emulators. So there are just too many ways a user can break or waste time customizing terminal and shell for execution. A drop down with presets is much better for now.

ouuan:
> Complicated because a typical detached execution is processed like this. CP Editor forks a Terminal, which in turn forks a shell (BASH) in this case but can be different if user sets those commands. Shell executes our program and then echos and waits for `read -n 1`, as pointed out earlier the `read` command depends on the shell `-n` is not available in all shells except `bash`. Then comes different syntax for maybe different terminal emulators. So there are just too many ways a user can break or waste time customizing terminal and shell for execution. A drop down with presets is much better for now.

What I want to let the user choose is only `konsole -e`, `gnome-terminal --`, etc.

translations/zh_CN.ts
```diff
@@ -657,8 +657,13 @@ git 提交编号: %3
         <translation>未能成功运行。请先编译。</translation>
     </message>
     <message>
-        <source>Please install xterm in order to use Detached Run.</source>
-        <translation>请安装 xterm 以使用在终端中运行。</translation>
+        <source>%1 could not be found. You can change the terminal from Preferences -&gt; Action -&gt; Detached Execution.</source>
```
IZOBRETATEL777, line 660: Please, update Chinese translations

ouuan:
> What I want to let the user choose is only `konsole -e`, `gnome-terminal --`, etc.

@coder3101 Any thoughts?

coder3101:
> > What I want to let the user choose is only `konsole -e`, `gnome-terminal --`, etc.
> 
> @coder3101 Any thoughts?

I accept the proposed idea. I will commit it now or you can commit it as I am not familiar with indentation for settings.json

coder3101:
@IZOBRETATEL777 Please update the Russian translations again. @ouuan Please review and add Chinese translations.

translations/zh_CN.ts
coder3101: He'll update it this time. I think he waited because he knew that some changes were to be brought.

src/Core/Runner.cpp
```diff
@@ -143,14 +126,36 @@ void Runner::runDetached(const QString &tmpFilePath, const QString &sourceFilePa
...
                                  " ^& pause\""));
     LOG_INFO("CMD Arguemnts " << runProcess->arguments().join(" "));
+#else
+    QString terminal = SettingsHelper::getTerminalProgram();
 
+    LOG_INFO("Using: " << terminal << " on Linux");
+    QProcess testProcess;
+    testProcess.start(terminal, {"-v"});
+    bool finished = testProcess.waitForFinished(2000);
+    if (!finished || testProcess.exitCode() != 0)
```
ouuan, line 143: Not all terminal emulators support `-v`, nor all terminal emulators support `--version`

It could be better if we handle the `errorOccurred` signal by checking whether the error is `QProcess::FailedToStart`. It is useful for both detached execution and normal execution.

src/Settings/settings.json
```diff
@@ -292,6 +292,18 @@
         "param": "QStringList {\"Light\", \"Dracula\", \"Monokai\", \"Solarized\", \"Solarized Dark\"}",
         "tip": "The syntax highlight theme of the code editor"
     },
+    {
+        "name": "Terminal Program",
+        "type": "QString",
+        "default": "xterm",
+        "tip": "The terminal emulator to use when running in detached mode.\nThis is usually the name or path of the terminal to use, \nSome possible values are konsole, gnome-terminal, xfce-terminal, xterm or any other terminal emulator program."
```
ouuan, line 299:
```suggestion
        "tip": "The terminal emulator to use when running in detached mode.\nThis is usually the name or the path of the terminal emulator.\nSome possible values are konsole, gnome-terminal, xfce-terminal, xterm or any other terminal emulator program."
```

src/Settings/settings.json
```diff
@@ -292,6 +292,18 @@
...
+    },
+    {
+        "name": "Terminal Arguments",
+        "type": "QString",
+        "default": "-e",
+        "tip": "Argument to pass to terminal emulator command to start execution of your program.\nThis is usually -e for most terminals like konsole, xterm, xfce-terminal but can be -- for gnome-terminal. Consult your terminal emulator program for the valid arguments."
```
ouuan, line 305:
```suggestion
        "tip": "Arguments used to execute a given command in the terminal emulator.\nThis is "-e" for most terminal emulators, including konsole, xterm, xfce-terminal but can be "--" for gnome-terminal.\nConsult your terminal emulator for the suitable arguments."
```

src/Settings/settings.json
```diff
@@ -292,6 +292,18 @@
         "param": "QStringList {\"Light\", \"Dracula\", \"Monokai\", \"Solarized\", \"Solarized Dark\"}",
         "tip": "The syntax highlight theme of the code editor"
     },
+    {
+        "name": "Terminal Program",
```
ouuan, line 296:
```suggestion
        "name": "Detached Run Terminal Program",
        "desc": “Terminal Program"
```

You should point out it's for detached run in the name.

If you want to omit "Detached Run" in the UI, you can set the "desc" attribute. However, I'd rather not to omit it.

coder3101: Isn't it implied the location of the setting.

Actions -> Detached Execution

"Terminal Program" will be detached terminal.

However, I accept the changes, Maybe we can add in future inbuild terminal and that will conflict with this.

src/Core/Runner.cpp
coder3101: This is out of the scope of this PR. We can open a new issue for this code refactoring. I suggest we can have a Util function like `isExecutable(QString path)`. The path is either name of program in which case it looks at environment or an absolute Path.

src/Settings/settings.json
ouuan: > Isn't it implied the location of the setting.
> 
> Actions -> Detached Execution

"Actions -> Detached Execution" is only in the preferences window, not with the setting itself. You don't see "Actions -> Detached Execution" when using `getTerminalProgram". And now the settings are independent of the preferences window.

src/Core/Runner.cpp
ouuan: > This is out of the scope of this PR.

It was fine with `xterm`, but now it breaks with some terminals. Are you sure we should fix this in another PR?

ouuan: BTW, (though this is out of the scope of this PR), I don't like the diagnostics for compiler and runner. Do you think it's very useful?

ouuan: What I want to change to is not a new way to check whether a program is available, but an error message when we fail to start a program.

coder3101:
I don't quite get it. But I think it's a small change and you can commit yourself.

coder3101:
@IZOBRETATEL777 We are waiting on your translations update to merge this PR.

coder3101:
@ouuan Need your approval.

PR #557, opened 2020-08-11
refactor: use SettingsHelper::getPath to get the path of a setting

ouuan:
## Description

Now we can use `SettingsHelper:getPathXXX` or `SettingsManager::getPathTextXXX` to get the path of a setting in the preferences window.

## Motivation and Context

This reduces duplicated translations, and prevent forgetting to change the path after moving a setting.

## How Has This Been Tested?

On Arch Linux.

## Type of changes

- [x] Refactor (changes which affect the meaning of the code but neither fix a bug nor add a feature)

### DIFF
--- a/src/Core/Checker.cpp
+++ b/src/Core/Checker.cpp
@@ -204,13 +204,14 @@ void Checker::onRunTimeout(int index)
 
 void Checker::onRunOutputLimitExceeded(int index, const QString &type)
 {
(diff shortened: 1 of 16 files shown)


coder3101:
What is parent in `SettingsHelper::pathofXYZ();`?

ouuan:
@IZOBRETATEL777 Please add Russian translations.

PR #559, opened 2020-08-12
chore(mainwindow):Use website instead of github url for problems parsed by competitive companion

coder3101:
This line was added prior to our website page. Website is our main entry point for new users to get themselves familiar to CP Editor. Moreover, Its short and easy to remember URL that github repository. Hence we now write "Powered by CP Editor (https://cpeditor.org)"

## Related Issue
N/A

## How Has This Been Tested?
Arch

## Screenshots (if appropriate)
N/A

## Type of changes

- [x] Chore (changes which do not belong to any type above)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/cpeditor/cpeditor/blob/master/CONTRIBUTING.md) document.
- [x] I have tested these changes locally, and this fixes the bug/the new feature behaves as the expectation.
- [x] The settings file in the old version can be used in the new version after this change.
- [x] These changes only fix a single bug/introduces a single feature. (Otherwise, open multiple Pull Requests instead, unless these bugs/features are closely related.)
- [x] The commit messages are clear and detailed. (Otherwise, use `git reset` and commit again, or use `git rebase -i` and `git commit --amend` to modify the commit messages.)
- [x] These changes don't remove an existing feature. (Otherwise, add an option to disable this feature instead, unless it's necessary to remove this feature.)
- [x] If there are changes of the text displayed in the UI, I have wrapped them in `tr()` or `QCoreApplication::translate()`.
- [x] I have documented these changes in [CHANGELOG.md](https://github.com/cpeditor/cpeditor/blob/master/doc/CHANGELOG.md), or these changes are not notable.

PR #561, opened 2020-08-13
feat(Competitive Companion): make head comments customizable

### LINKED ISSUE #203: customize head comments

More flexible head comments (like adding current time and so on).

Set some identifier. When the template was loaded, replace the identifier (like the template of cf-tools).

N/A

**Additional context**

ouuan:
## Description

Now the head comments for Competitive Companion is customizable.

## Related Issue

This closes #203.

This takes over #559.

## How Has This Been Tested?

On Arch Linux.

## Type of changes

## Checklist

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -512,7 +512,7 @@ void MainWindow::applyCompanion(const Extensions::CompanionData &data)
                 value = value[attr];
             if (value.isUndefined())
             {
-                log->error(tr("Companion"),
+                log->error("Companion",
                            tr("Unknown attribute: [%1]. Please check the head comments setting at %2.")
                                .arg(path.join('.').arg(SettingsHelper::pathOfCompetitiveCompanionHeadComments())));
             }
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -495,7 +495,8 @@ void MainWindow::applyCompanion(const Extensions::CompanionData &data)
     if (isUntitled() && !isTextChanged())
     {
         QString comments = SettingsHelper::getCompetitiveCompanionHeadComments().replace(
-            "${time}", QDateTime::currentDateTime().toString("yyyy-MM-dd HH:mm:ss"));
+            "${time}",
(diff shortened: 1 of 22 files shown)


src/mainwindow.cpp
```diff
@@ -494,21 +494,53 @@ void MainWindow::applyCompanion(const Extensions::CompanionData &data)
 
     if (isUntitled() && !isTextChanged())
     {
-        QString meta = data.toMetaString();
-        meta.prepend("\n");
-        meta.append("Powered by CP Editor (https://github.com/cpeditor/cpeditor)");
+        QString comments = SettingsHelper::getCompetitiveCompanionHeadComments().replace(
+            "${time}", QDateTime::currentDateTime().toString("yyyy-MM-dd HH:mm:ss"));
```
coder3101, line 498: You can also split it into individual $time and $date.

src/mainwindow.cpp
```diff
@@ -494,21 +494,53 @@ void MainWindow::applyCompanion(const Extensions::CompanionData &data)
...
+        {
+            auto match = it.next();
+            finalComments += comments.mid(lastEnd, match.capturedStart() - lastEnd);
+            auto path = match.captured().mid(7, match.capturedLength() - 8).split(".");
+            auto value = QJsonValue::fromVariant(data.doc.toVariant());
+            for (auto attr : path)
+                value = value[attr];
+            if (value.isUndefined())
+            {
+                log->error(tr("Companion"),
```
coder3101, line 515: Competitive Companion is a name of a software product. Does it needs to be translated?

src/mainwindow.cpp
ouuan: What about let the user set the format by `yyyy-MM-dd HH:mm:ss`?

coder3101: There are 2 options:
1. You need to add a new Preference page called "Header Comments" and let the user customize everything about Header comments.
2. Add placeholders for individual parts of date time like. $month $date $day $year $minute etc...
Whichever you feel is better.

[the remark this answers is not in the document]
ouuan: The setting is separated for each user, so it's easy to directly set the name instead of using the username of the computer/OS.

PR #565, opened 2020-08-15
feat: add features for solving grader problems

ouuan:
## Description

1.  Now the working directory of the compiler and the document path for the language server are the source file path, and `-I <source file path>` is added in the C++ compile command, so that you can easily include headers and compile with the grader, using relative paths and get the correct linting.
2.  Now you can set custom compile command and time limit for each tab.

## Motivation and Context

I took part in [APIO](https://apio2020.id/) today, and found that it's not very easy to solve grader problems with CP Editor (I used `#ifdef OUUAN` during the contest, and the biggest problem was the language server). So these features are added to make it easier to solve grader problems.

The custom compile command and time limit are also helpful without grader problems. For example, you can use different language standards for different problems, or toggle O2 in a single tab, etc.

## How Has This Been Tested?

On Arch Linux with the problems of APIO2020.

## Type of changes

## Checklist

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -564,6 +564,9 @@ void MainWindow::applyCompanion(const Extensions::CompanionData &data)
         testcases->addTestCase(data.testcases[i].input, data.testcases[i].output);
 
     setProblemURL(data.url);
+
+    if (SettingsHelper::isCompetitiveCompanionSetTimeLimitForTab())
+        customTimeLimit = data.timeLimit;
 }
 
 void MainWindow::applySettings(const QString &pagePath, bool shouldPerformDigonistic)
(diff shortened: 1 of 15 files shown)


src/mainwindow.cpp
```diff
@@ -211,7 +210,7 @@ void MainWindow::run(int index)
     connect(tmp, SIGNAL(runKilled(int)), this, SLOT(onRunKilled(int)));
     tmp->run(tmpPath(), filePath, language, SettingsManager::get(QString("%1/Run Command").arg(language)).toString(),
              SettingsManager::get(QString("%1/Run Arguments").arg(language)).toString(), testcases->input(index),
-             SettingsHelper::getTimeLimit());
```
coder3101, line 214: Shouldn't we update the tooltip and or this Setting to "Default Time limit"

src/mainwindow.cpp
```diff
@@ -211,7 +210,7 @@ void MainWindow::run(int index)
...
+             timeLimit());
```
coder3101, line 213: Now that every tab has its own timelimit. We can add another verdict "TLE" and also pull timelimit for the tab from Competitive Companion

ouuan: > We can add another verdict "TLE"

That worths another PR. Though different time limits for different tabs make the TLE verdict more useful, it's not really related to this. i.e. this PR works without the TLE verdict and the TLE verdict without this PR. I can open another PR for the TLE verdict without conflicts with this PR.

PR #516, opened 2020-07-22
feat: use KSyntaxHighlighting and rewrite the code editor with vim mode

coder3101:
Use Kate's Syntax highlighter engine for Code Highlighting.

## Description
`QCodeEditor` uses very basic RegEx based matching to highlight code. This has many issues and they are listed in related issues section in the side.

**ECM and Syntax highlighting builds and links to a shared library**.

## Related Issue
Fixes #174 and Fixes #210 and Fixes #533 and Fixed #372 and Fixes #393 and Fixes #413

## Motivation and Context
To make syntax highlighting more better.

## How Has This Been Tested?
Tested on Windows 10

## Screenshots (if appropriate)
image

## Type of changes

- [x] Bug fix (changes which fix an issue)
- [x] New feature (changes which add functionality)
- [ ] Documentation (changes which modify the documentation only)
- [ ] Style change (changes which do not affect the meaning of the code: code formatting, etc.)
- [ ] Refactor (changes which affect the meaning of the code but neither fix a bug nor add a feature)
- [ ] Performance improve (changes which improve performance)
- [ ] Test (changes which add tests)
- [x] Build (Changes that affect the build system or external dependencies)
- [x] CI (changes to CI configuration files and scripts)
- [ ] Chore (changes which do not belong to any type above)
- [ ] Revert (revert previous changes)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/cpeditor/cpeditor/blob/master/CONTRIBUTING.md) document.
- [x] I have tested these changes locally, and this fixes the bug/the new feature behaves as the expectation.
- [x] The settings file in the old version can be used in the new version after this change.
- [x] These changes only fix a single bug/introduces a single feature. (Otherwise, open multiple Pull Requests instead, unless these bugs/features are closely related.)
- [ ] The commit messages are clear and detailed. (Otherwise, use `git reset` and commit again, or use `git rebase -i` and `git commit --amend` to modify the commit messages.)
- [ ] These changes don't remove an existing feature. (Otherwise, add an option to disable this feature instead, unless it's necessary to remove this feature.)
- [ ] If there are changes of the text displayed in the UI, I have wrapped them in `tr()` or `QCoreApplication::translate()`.
- [ ] I have documented these changes in [CHANGELOG.md](https://github.com/cpeditor/cpeditor/blob/master/doc/CHANGELOG.md), or these changes are not notable.

## Extra Checklist

- [ ] Make bottom margin work correctly.
- [ ] Show squiggles in the sidebar.
[...]

Issue #214, opened 2020-03-29
Make most shortcuts customizable


ouuan:
**Is your feature request related to a problem? Please describe.**

Now only a few shortcuts are customizable.

**Describe the solution you'd like**

After #166 is done, it will be possible to make most shortcuts customizable and override the default shortcuts.

**Describe alternatives you've considered**

N/A

**Additional context**

The shortcuts that won't be customizable include some Qt shortcuts like <kbd>Ctlr+Tab</kbd>/<kbd>Ctrl+Shift+Tab</kbd> to switch between tabs, `QTextEdit` shortcuts, and <kbd>Tab</kbd>(with a selection)/<kbd>Shift+Tab</kbd> for indentation.

cs-mshah:
additional shortcuts like ctrl+(+/-) for increasing decreasing font size and many other shortcuts from other editors like sublime text. This editor can have shortcuts for everything like 'submit' and can be made so that minimum or no use of mouse is required.

cs-mshah:
@ouuan Could you add shortcuts like crtrl+D to select next occurrence of the selected word. This is **REALLY NEEDED** to change something like datatypes form ll to int or vice versa.  
Currently ctrl+D is having some funny effect and it deletes the selected word.

ouuan:
> @ouuan Could you add shortcuts like crtrl+D to select next occurrence of the selected word. This is **REALLY NEEDED** to change something like datatypes form ll to int or vice versa.
> Currently ctrl+D is having some funny effect and it deletes the selected word.

I don't see it's related to this thread. This thread is for making shortcuts for existing actions customizable, not for creating new actions and binding shortcuts for them.

For the use case mentioned in your request, why don't you use Find & Replace?

fengwei2002:
Is it possible to make a switch that allows the user to use three different shortcuts to switch between three different views?

For example, when half the screen is used, you can use 1 2 instead of 3
When using full screen use 1 3 instead of 2

What I mean is that we can use three different shortcut keys to quickly switch from one layout to another layout, without having to combine the switching shortcut keys of the three layouts

huaruoji:
Yes, I would very much like to be able to delete a line with ctrl+d

PR #581, opened 2020-08-31
feat(Editor): Add Vim mode

### LINKED ISSUE #210: Better highlighting

It is harder to see highlight characters in dark theme.

Maybe make it clear like this?
Also an option to remove highlight at current line might help

### LINKED ISSUE #220: Please add Vim Mode！！！
It's such a wonderful editor. Could you add vim mode in it?

### LINKED ISSUE #533: Solarized Dark focus area is not readable
While using cpeditor, I found out that the solarized dark theme is unusable due to the color used when the line is in focus. The screenshot would provide more info. I am comparing the cpeditor solarized dark theme with vscode solarized dark theme.

Make it like in vscode.

image

image

coder3101:
Vi mode is very popular and is widely used by many programmers. It speeds up so many tasks and so those who are used to it, do not like to code on any other editor. This is stopping many from adapting CP Editor as their main Editor for competitive Programming. This PR adds support for Vi mode in the code Editor using a popular extension written by the authors of Qt Creator called FakeVim. 

## Motivation and Context
I like to code in Vim too and sometimes I don't like CP Editor because of it not having Vim Support.

## How Has This Been Tested?
- Primarily testing and being developed on Arch Linux
- Final testing will include Windows 10

N/A

## Type of changes

## Checklist

# Checklist 
- [ ] Switch to master branch for submodule QCodeEditor. https://github.com/cpeditor/QCodeEditor/pull/28 should be merged
- [ ] Associated docs PR (https://github.com/cpeditor/cpeditor.github.io/pull/115) should be merged to website.


coder3101:
@ouuan Please review the PR or list down the issues, So I can acknowledge them or point if those could be solved.

ouuan:
Some problems:

1.  Many shortcuts are not available, including Ctrl+O, Ctrl+P, Ctrl+T, Ctrl+K, etc. Maybe #214 helps.
2.  Fake Vim doesn't use the code editor settings. The proper way to do this is to use a `.vimrc`, so I suggest moving the Fake Vim settings to the Code Edit section, and add "Fake Vim is not enabled" as a dependency of other code editor settings:
    ```json
        "depends": [
            {
                "name": "FakeVim/Enable",
                "check": "return !var.toBool();"
            }
        ],
    ```
3.  https://github.com/cpeditor/QCodeEditor/pull/28#issuecomment-683399356
4.  When the cursor at the beginning/end of a line, and move to a character left/right, the current line highlight disappears.

src/Extensions/FakeVimProxy.cpp
```diff
@@ -0,0 +1,506 @@
...
+{
+    handler->handleCommand(_("set nopasskeys"));
+    handler->handleCommand(_("set nopasscontrolkey"));
+
+    // Set some Vim options.
+    handler->handleCommand(_("set expandtab"));
+    handler->handleCommand(_("set shiftwidth=8"));
+    handler->handleCommand(_("set tabstop=%1").arg(SettingsHelper::getTabWidth()));
+    handler->handleCommand(_("set autoindent"));
+    handler->handleCommand(_("set smartindent"));
```
ouuan, line 453: Maybe some code editor settings can be set here? At least, now the tab width is not set properly. I'm wondering why `set shiftwidth=8`.

src/Extensions/FakeVimProxy.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+    handler->handleCommand("source " + SettingsHelper::getFakeVimRCLocation());
```
ouuan, line 455: Check the location exists, or is not empty first.

ouuan (changes requested):

src/Extensions/FakeVimProxy.cpp
coder3101: Yes, that's why I want to make a separate settings page for FakeVim.

src/Extensions/FakeVimProxy.cpp
coder3101: Thanks, forgot it!

src/Extensions/FakeVimProxy.cpp
coder3101: Shift width and tab width are explained [here](https://www.reddit.com/r/vim/comments/99ylz8/confused_about_the_difference_between_tabstop_and/)

ouuan: > Yes, that's why I want to make a separate settings page for FakeVim.

I don't get it. I meant if you want, you can apply some settings in the Code Edit section to FakeVim. It can't be the reason you want to make a separate page.

Do you want to add a convenient way to configure Fake Vim besides writing a `.vimrc`?

coder3101: Initially I was thinking of it but if you say, I can reuse existing Code Edit options like Tab Width. Replace with Space (expandtabs) etc.

ouuan: Basically, what I want is, if a setting in the Code Edit section is not used by Fake Vim, it should be disabled (i.e. it depends on Fake Vim is not enabled).

ouuan: And, if there are only the "Enable" and "vimrc" settings, I think it should in the Code Edit page. At least, it's not an extension. I thought an extension is a third party software not included in CP Editor itself.

src/Settings/settings.json
```diff
@@ -1066,5 +1066,26 @@
         "type": "int",
         "default": 5,
         "notr": true
+    },
+    {
+        "name": "FakeVim/Enable",
+        "desc": "Enable FakeVim emulation",
+	"type": "bool",
+	"tip": "Enable vim Emulation in Code Editor",
+        "notr": true
```
ouuan, line 1075: Why notr?

src/Settings/settings.json
```diff
@@ -1066,5 +1066,26 @@
...
+    },
+    {
+        "name": "FakeVim/RCLocation",
+	"desc": "Vim RC Location",
+        "type": "QString",
+	"ui": "PathItem",
+        "tip":"Path to your Vim RC file",
+        "notr": true,
```
ouuan, line 1083: Why notr?

ouuan (changes requested):

src/Extensions/FakeVimProxy.cpp
coder3101: I am thinking that instead of loading it from File We can provide a `QPlainTextEdit` where user puts their config. Like Clang Format Config. The benefit would be that it will be more portable, (configuration will be exported along with rest of settings). It is because usually FakeVim does not supports full Vim Commands, plus in most of users vim configs the configs for VimPlug/Pathogen are present which is completely unsupported in FakeVim.

src/Extensions/FakeVimProxy.cpp
coder3101: Moved to Core and Settings moved to Code Edit

src/Core/FakeVimProxy.cpp
```diff
@@ -0,0 +1,494 @@
...
+}
+
+QString FakeVimProxy::content() const
+{
+    return document()->toPlainText();
+}
+
+void FakeVimProxy::initHandler(FakeVim::Internal::FakeVimHandler *handler)
+{
+    QStringList lines = SettingsHelper::getFakeVimRC().split(QRegExp("[\r\n]"), Qt::SkipEmptyParts);
```
ouuan, line 444: https://stackoverflow.com/questions/10831565/how-do-i-split-long-lines-in-a-vimrc-file

src/Core/FakeVimProxy.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+void FakeVimProxy::initHandler(FakeVim::Internal::FakeVimHandler *handler)
+{
+    QTemporaryFile file;
+    if (file.open())
+    {
+        file.write(SettingsHelper::getFakeVimRC().toLocal8Bit());
+        file.close();
+        handler->handleCommand("source " + file.fileName());
+    }
```
ouuan, line 446: Else error.

coder3101:
>     1. Many shortcuts are not available, including Ctrl+O, Ctrl+P, Ctrl+T, Ctrl+K, etc. Maybe #214 helps.

I was thinking that since mostly vim users configure their commands to automate stuff like testing, I was thinking of adding special vim commands that does all those what above actions. For example, to detach run you can put `:drun` in command, to open `:open` and similarly. We can even add a Preference page for use to choose their own commands. These will be handled by us in FakeVim Proxy. 

The reason is that if we somehow make FakeVim pass Control keys, then some FakeVim features will not work, like Ctrl+V (Block Visual) and if we make FakeVim intercept Control keys, then Ctrl+P is intercepted by FakeVim and not passed to Application. Settings a filter that conditionally chooses to pass keys to application and editor is a bad idea too because user can customize any key combinations in vimrc. So it gets much complicated. 

#214 is something that I think would only make sense in non-vim mode.

src/Core/FakeVimProxy.cpp
coder3101: I log it in LOG_ERR. Should I pass complete window and show error in message logger.

coder3101:
>     4\. When the cursor is at the beginning/end of a line, and moved to a character left/right, the current line highlight disappears.

I tried to investigate this issue and found that it is also related to Vim behaviour.  Highlighting of the current line is done by the `highlightCurrentLine()` slot which is signalled upon `QTextEdit::cursorPositionChanged()`. This makes sense when cursor can move to next line if it is at end of a line and a right arrow key is pressed or cursor moves to last line if its at the beginning of a line and left arrow is pressed. In otherwords, any navigation key press is sure to change the cursor position. However in FakeVim, this is not always true. If you are at end of a line and press `l` (to move right) the cursor does not moves at all. It stays at its location and hence the signal is never emitted and so the highlighing is not done but the last `setExtraSelection()` is automatically cleared by Qt upon a keypress.

At this point, the best approach I can think of is that we disable current Line highlighing for vim mode. By default vi/vim/nvim none of them have the current line highlighting feature. As we working on our own new Code Editor we can override the Code Editor's `cursorPositionChanged` to emit it when in end/start of line in vim mode.

coder3101:
> > ```
> > 1. Many shortcuts are not available, including Ctrl+O, Ctrl+P, Ctrl+T, Ctrl+K, etc. Maybe #214 helps.
> > ```
> 
> I was thinking that since mostly vim users configure their commands to automate stuff like testing, I was thinking of adding special vim commands that does all those what above actions. For example, to detach run you can put `:drun` in command, to open `:open` and similarly. We can even add a Preference page for use to choose their own commands. These will be handled by us in FakeVim Proxy.
> 
> The reason is that if we somehow make FakeVim pass Control keys, then some FakeVim features will not work, like Ctrl+V (Block Visual) and if we make FakeVim intercept Control keys, then Ctrl+P is intercepted by FakeVim and not passed to Application. Settings a filter that conditionally chooses to pass keys to application and editor is a bad idea too because user can customize any key combinations in vimrc. So it gets much complicated.
> 
> #214 is something that I think would only make sense in non-vim mode.

@ouuan I want your opinion on this before I start working on it.

coder3101:
I'll start to work on it anyways.

coder3101:
There are 2 things left before this PR can be ready for review:
- Add /  Implement commands that require arguments.
- When a custom command is invoked, we should show some feedback in the status bar. Like when `:drun` we can show "Detached Window has opened"

coder3101:
```
/home/ashar/Projects/cpeditor/src/Core/FakeVimCommand.cpp:117: Qualifying with unknown namespace/class ::FakeVimCommand
/home/ashar/Projects/cpeditor/src/Core/FakeVimCommand.cpp:134: Qualifying with unknown namespace/class ::FakeVimCommand
```

I am getting these warnings when updating translations. Regardless the translations are being correctly updated.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include <QFileInfo>
+#include <QStatusBar>
+
+namespace Core
+{
+FakeVimCommand::FakeVimCommand(AppWindow *aw) : appwin(aw)
+{
+}
+
+bool FakeVimCommand::handle(FakeVim::Internal::ExCommand const &cmd)
```
coder3101, line 33: This function handles the custom commands. It calls many helper methods like `wantSomething(cmd)` which returns true. If command `cmd` is requesting to do what the function names say. In some cases like `open` the function returns a `QPair<bool, QString>`, bool indicates that `cmd` matched with the action represented by function name, and if it is true, the QString contains the arguments to the command. The arguments are verified beforehand, so in this case, the open ensures that file to open exists and is a c/c++/java/py source. If the argument failed the checks, an empty string is returned and the `wantSomething(cmd)` has warned the user of why it failed to return the argument.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,266 @@
...
+            return {true, 0};
+        }
+        bool ok = false;
+        int caseNum = ex.args.toInt(&ok);
+        if (ok)
+            return {true, caseNum};
+        else
+        {
+            if (appwin->currentWindow())
+                appwin->currentWindow()->statusBar()->showMessage(tr("%1 is not a number").arg(caseNum),
```
coder3101, line 225: These error messages will timeout after 2 seconds and restore the status bar to its older state.

src/Core/FakeVimProxy.cpp
```diff
@@ -0,0 +1,491 @@
+/*
```
coder3101, line 1: Almost 95% of this file is taken from FakeVim Examples with some improvements.

src/Core/FakeVimProxy.cpp
```diff
@@ -0,0 +1,491 @@
...
+#include <QTextBlock>
+#include <QTextEdit>
+#include <QTextStream>
+#include <QWidget>
+
+namespace Core
+{
+
+using _ = QLatin1String;
+FakeVimProxy::FakeVimProxy(QWidget *widget, MainWindow *mw, AppWindow *aw, QObject *parent)
```
coder3101, line 58: FakeVim Proxy takes an AppWindow instance to invoke actions slots like that which opens preference window, new file, format and others.

src/Settings/PreferencesWindow.cpp
```diff
@@ -166,9 +166,9 @@ PreferencesWindow::PreferencesWindow(QWidget *parent) : QMainWindow(parent)
 #define TRKEY(x) x, tr(x)
 
     AddPageHelper(this)
-        .page(TRKEY("Code Edit"),
+        .page(TRKEY("Code Editing"),
```
coder3101, line 169: This was looking incomplete. I have renamed it into something more complete.

src/appwindow.cpp
```diff
@@ -345,11 +345,11 @@ void AppWindow::maybeSetHotkeys()
     }
 }
 
-bool AppWindow::closeTab(int index)
+bool AppWindow::closeTab(int index, bool noConfirmQuit = false)
```
coder3101, line 348: In order to handle `q!` (forcefully close). We should have a way to close the window without the confirmation. This argument has been added to handle that.

src/appwindow.cpp
```diff
@@ -386,7 +387,7 @@ void AppWindow::openTab(MainWindow *window)
     onEditorFileChanged();
 }
 
-void AppWindow::openTab(const QString &path)
+void AppWindow::openTab(const QString &path, const QString &lang)
```
coder3101, line 390: To provide `new py` command that opens a new tab with the given language. openTab should be able to open a Tab with the language specified. If it is empty, the lang is detected from path suffix.

src/appwindow.cpp
```diff
@@ -1328,6 +1332,13 @@ void AppWindow::onConfirmTriggered(MainWindow *widget)
         ui->tabWidget->setCurrentIndex(index);
 }
 
+void AppWindow::onWindowCloseRequested(MainWindow *window)
```
coder3101, line 1335: To actually close the window with `:q`, mainwindow must be able to request appwindow to close itself.

coder3101:
@ouuan I have placed some inline comments to explain what I did and why. You can review them and propose changes if you want.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    if (ex.cmd == "vmode" || ex.cmd == "vmd")
+    {
+        if (ex.args == "split" || ex.args == "edit")
+            return {true, ex.args};
+        else
+        {
+            if (appwin->currentWindow())
+                appwin->currentWindow()->statusBar()->showMessage(
+                    tr("%1 is not a valid view mode. It should be one of \"split\", \"edit\"").arg(ex.args),
```
coder3101, line 264: User cannot switch to IO mode because, in IO mode the editor is hidden and hence vim commands cannot be issued.

ouuan:
You can put some comments in the code.

coder3101:
> You can put some comments in the code.

Sure, I will put them. Wait for new commits.

coder3101:
Opened PR for English docs at https://github.com/cpeditor/cpeditor.github.io/pull/84

coder3101:
@IZOBRETATEL777 this PR is also almost ready for translation.

IZOBRETATEL777:
> @IZOBRETATEL777 this PR is also almost ready for translation.

coder3101:
> Should I remove awaiting for translations label?

Yes you should remove it on all PR that you add translations to.

src/mainwindow.cpp
```diff
@@ -30,6 +30,7 @@
 #include "Util/FileUtil.hpp"
 #include "Util/QCodeEditorUtil.hpp"
 #include "Widgets/TestCases.hpp"
+#include "appwindow.hpp"
```
ouuan, line 33:
```suggestion
```

I can compile without this.

src/Core/FakeVimProxy.hpp
```diff
@@ -0,0 +1,102 @@
...
+    void setStatusBar();
+    void invalidate();
+    bool hasChanges();
+    QTextDocument *document() const;
+    QString content() const;
+
+    QWidget *m_widget;
+    MainWindow *m_mainWindow;
+    AppWindow *m_appWindow;
+    FakeVimCommand m_commandHandler;
```
ouuan, line 93: Why not pointer?

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include <QFileInfo>
+#include <QStatusBar>
+
+#define SHOW_STATUS_MESSAGE(x)                                                                                         \
+    if (appwin->currentWindow())                                                                                       \
+        appwin->currentWindow()->statusBar()->showMessage(x, STATUS_RESPONSE_TIMEOUT);
+
+namespace Core
+{
+FakeVimCommand::FakeVimCommand(AppWindow *aw) : appwin(aw)
```
ouuan, line 33:
```suggestion
FakeVimCommand::FakeVimCommand(AppWindow *aw) : appwin(aw), QObject(aw)
```

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include "appwindow.hpp"
+#include "fakevimhandler.h"
+#include "generated/SettingsHelper.hpp"
+#include <QApplication>
+#include <QFileInfo>
+#include <QStatusBar>
+
+#define SHOW_STATUS_MESSAGE(x)                                                                                         \
+    if (appwin->currentWindow())                                                                                       \
+        appwin->currentWindow()->statusBar()->showMessage(x, STATUS_RESPONSE_TIMEOUT);
```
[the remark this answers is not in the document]
ouuan: It seems that it always shows an error, you can call the function `showError` and remove some duplicate text "Error".

ouuan (changes requested):

ouuan:
Is there a command to switch between tabs?

CHANGELOG.md
```diff
@@ -1,7 +1,14 @@
 ## UNRELEASED
 
+Vim Emulation has been added to the Code Editor. Almost all vim commands are supported, a complete list of supported commands could be found [here](https://github.com/cpeditor/FakeVim#supported-features). In vim, Control key is widely used in normal mode, In order to correctly emulate vim, all Control Keys are intercepted by Code Editor and not by the application resulting in some shortcuts not working. Specifically, any CP Editor shortcut that starts and `Ctrl+X+Y`, `X` being any key except `Alt` or `Shift` will be intercepted by Code Editor. Hence **popular shortcuts like `Ctrl+O`, `Ctrl+N` will not work, if Vim mode is enabled**.
+
+In order to compensate for the loss of Control key shortcut in Vim Emulation, we have added some custom commands which can perform various task in CP Editor. These commands are faster and mostly preferred by Vim users. For Example, To open a new tab you can issue `open` or `opn` (shorthand) command. To run the third testcase you can issue `run 3` command. The complete list of all the custom commands could be found [here](https://cpeditor.org/docs/preferences/general/#custom-vim-commands)
```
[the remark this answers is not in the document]
coder3101: You can suggest a short version.

src/Settings/settings.json
```diff
@@ -429,7 +447,25 @@
...
+        "name": "Cursor Overwrite",
+        "desc": "Default cursor overwrites",
+        "type": "bool",
+        "tip": "Default cursor is the cursor you see on opening new tabs.\nIf enabled, default cursor will overwrite next character.\nThis behaviour can be toggled for each tab by pressing INS/Insert on your keyboard.",
+        "depends": [
+            {
+                "name": "FakeVim/Enable",
+                "check": "return !var.toBool();"
+            }
+        ]
```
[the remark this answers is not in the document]
coder3101: This is something not related with fakevim.

Cursor in normal editing is of two forms, INSERT and REPLACE. In insert mode cursor is between characters and new characters is inserted in between.

In REPLACE mode, cursor is always over a character and new character overwrites the character over which cursor is present.

This setting tells when new tabs are opened in which mode the cursor should be in non-vim mode? By default it is in INSERT mode. For those who like to code in REPLACE mode will have to press `INS` to change cursor mode everytime new tabs are opened. This setting just helps them in it.

The behaviour of toggling edit mode was not present in QCodeEditor prior to this PR. So you can check the related QCodeEditor PR.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,288 @@
...
+            SHOW_STATUS_MESSAGE(tr("Error: %1 requires a file path").arg(ex.cmd));
+            return {true, ""};
+        }
+
+        QFileInfo file(ex.args);
+        if (!file.exists())
+        {
+            SHOW_STATUS_MESSAGE(tr("Error: %1 does not exists").arg(ex.args));
+            return {true, ""};
+        }
```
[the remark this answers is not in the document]
coder3101: > Actually CP Editor can open a tab with a non-exist file path.

`open` and `new` are two different commands. One is used to edit existing files and hence requires file path to exist. `new` is like opening new buffer. Doesn't need to have file path.

How do you open a file that doesn't exists with `Ctrl+O`?

src/Core/FakeVimCommand.cpp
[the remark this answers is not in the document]
coder3101: I'll do that refactoring. Thanks for pointing it out.

src/Core/FakeVimCommand.cpp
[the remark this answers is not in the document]
ouuan: This behavior is usually done by running `cpeditor <a non-exist path>` in the command line. It's a quick way to set the file name when creating a new tab. It's not available via Ctrl+O because QFileDialog only allows existing files.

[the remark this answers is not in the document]
ouuan: This is documented [here](https://cpeditor.org/docs/tips/#launch-cp-editor-in-the-command-line).

coder3101:
> Is there a command to switch between tabs?

We can use `:bnext`, `:bprev` and `:b n` (n is a number) to switch next, last, nth tab respectively.

The only problem is that in changing tab the status bar also changes. Whereas in actual vim the status bar and buffer(text area) are two different things and status bar doesn't changes.

I am thinking of using `appwindows` statusbar instead of `mainwindow`. I'll try that and let you know.

src/Core/FakeVimCommand.cpp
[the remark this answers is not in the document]
coder3101: `open` command emulates `Ctrl+O` also vim equivalent for open is `:e` which opens only existing file paths. 

I never tried  `cpeditor <path>` but it seems like it creates a path and opens in new tabs. (Open should not create files)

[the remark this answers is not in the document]
ouuan: 1. Since `open` emulates Ctrl+O, I think it can open a file dialog when arguments are empty.
2. No file is created when opening a non-exist file. It's simply a quick way to set the path used when saving it. It's the same as `vim <non-exist path>`.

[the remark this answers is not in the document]
coder3101: >  No file is created when opening a non-exist file. It's simply a quick way to set the path used when saving it. It's the same as `vim <non-exist path>`

This is a good approach for non existing paths. So how exactly do we set such path in new tabs?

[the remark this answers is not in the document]
ouuan: > So how exactly do we set such path in new tabs?

Nothing different from an existing path.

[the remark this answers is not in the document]
coder3101: I think default behaviour of open should be to error out in non-existing paths. However, I can add handling `!` (bang) command so that it forcefully opens a path. An example would be `open! <non-existing-path>`

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,288 @@
...
+        {
+            return {true, 0};
+        }
+        bool ok = false;
+        int caseNum = ex.args.toInt(&ok);
+        if (ok)
+            return {true, caseNum};
+        else
+        {
+            SHOW_STATUS_MESSAGE(tr("%1 is not a number").arg(caseNum));
```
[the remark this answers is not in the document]
coder3101: I will solve it by making fakevimhandler shareable across different tabs. So YANK in one tab can PASTE in other and so on and so forth. It also needs some code refactoring on which I will solve this.

src/Core/FakeVimProxy.hpp
coder3101: Yes, that seems a better idea. the handler shouldn't be created and destroyed often. It can be reused if allocated on heap.

src/Core/FakeVimCommand.cpp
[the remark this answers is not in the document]
coder3101: > YANK in one tab can PASTE in other and so on and so forth.

This can already be achieved even if we use multiple different handlers. They all share the same YANK history.

As for the commands, I am refactoring the FakeVimCommands file completely.

coder3101:
@IZOBRETATEL777 please update the translations.

coder3101:
> Is there a command to switch between tabs?

You can switch to next tab with command `next` and `previous` will switch to the last tab. They cycle like a chain.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,263 @@
...
+    case CommandTypes::OPEN: {
+        QString path = args;
+        if (path.isEmpty())
+        {
+            showError(tr("open requires a file path"));
+            break;
+        }
+
+        if (path.startsWith('~'))
+            path = QDir::home().absolutePath() + "/" + args.mid(1);
```
ouuan, line 143:
```suggestion
            path = QDir::home().filePath(args.mid(1));
```

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+        QFileInfo file(path);
+        if (file.isRelative())
+            file.makeAbsolute();
```
ouuan, line 147: You were still using `absoluteFilePath` in other places without a single `filePath` so this was useless.

And the `file.isRelative()` check is not necessary.

BTW, the path is converted to canonical form in `MainWindow::setFilePath`, so you don't need to convert the path to open a file.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,263 @@
...
+            !Util::javaSuffix.contains(file.suffix()))
+        {
+            showError(tr("%1 is not c++, python or java source file").arg(file.absoluteFilePath()));
+            break;
+        }
+
+        if (!file.exists() && !hasbang)
+        {
+            showError(
+                tr("%1 does not exists. To open non-existing file use open! instead").arg(file.absoluteFilePath()));
```
ouuan, line 159:
```suggestion
                tr("[%1] does not exists. To open a tab with a non-existing file, use [open!] instead").arg(path));
```

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,263 @@
...
+        if (last != curr)
+            appwin->setTabAt(last);
+        break;
+    }
+
+    case CommandTypes::OPEN: {
+        QString path = args;
+        if (path.isEmpty())
+        {
+            showError(tr("open requires a file path"));
```
ouuan, line 138: I suggest opening the file dialog here.

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+        // Parsed index correctly
+        if (ans == 0)
+            appwin->on_actionRun_triggered();
+        else if (ans > 0 && appwin->currentWindow() && appwin->currentWindow()->testcases &&
+                 appwin->currentWindow()->testcases->count() >= ans)
+            appwin->currentWindow()->runTestCase(ans - 1);
+        else
+            showError(tr("cannot run testcase %1").arg(args));
+        break;
```
ouuan, line 196: I think the logic could be clearer:

1. If `args.isEmpty()`, do something and break.
2. If the argument is not a number, show error and break.
3. If the case number is valid, do something and break.
4. Do something.

Instead of:

1. `ans = -1`.
2. If `args.isEmpty()`, `ans = 0`.
3. Else if the case number is a number, `ans = caseNum`; otherwise, show error and break.
4. If `ans == 0`, do something.
5. If `ans > 0` and `ans` is valid, do something.
6. Else do something.

src/Core/FakeVimCommand.hpp
```diff
@@ -0,0 +1,89 @@
...
+        KILL_PROCESS,
+        FORMAT_CODE,
+        SNIPPETS,
+        VMODE,
+        PREFERENCE,
+        CHLANG,
+        CLEAR,
+        EXIT,
+
+        UNKNOWN
```
ouuan, line 60: Why not CamelCase?

ouuan (changes requested):
I'm wondering the purpose of splitting `customCommandType` and `handleCustomCommand` into two functions :thinking:

src/Core/FakeVimCommand.cpp
coder3101: `file.suffix()` and `file.absolutePath()` are used so QFileInfo

src/Core/FakeVimCommand.cpp
coder3101: This will be same as clicking open in action menu

src/Core/FakeVimCommand.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    case CommandTypes::Open: {
+        QString path = args;
+
+        if (path.isEmpty())
+        {
+            auto fileNames = DefaultPathManager::getOpenFileNames(
+                "Open File", appwin->currentWindow(), tr("Open Files"), Util::fileNameFilter(true, true, true));
+            appwin->openTabs(fileNames);
```
ouuan, line 142: `appwin->on_actionOpen_triggered();`

src/Core/FakeVimCommand.cpp
coder3101:
```
1. if(args is empty) do stuff
2. else 
       2.1 if(args is not parseable to int) error nan
       2.2 else 
               2.2.1 if (args is in range) runOnly this testcase
               2.2.2 else error not a valid number.  
```

I did it like this, no breaks

src/Core/FakeVimCommand.hpp
coder3101: Sure, Why not

coder3101:
> I'm wondering the purpose of splitting `customCommandType` and `handleCustomCommand` into two functions
In future, if we will to offer a way to customize the name of custom commands, the implementation of a command and the check for command name should be clean. 

Anyways
What do you suggest?

ouuan:
> In future, if we will to offer a way to customize the name of custom commands, the implementation of a command and the check for command name should be clean.

OK.

src/Core/FakeVimProxy.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+bool FakeVimProxy::wantQuit(const FakeVim::Internal::ExCommand &cmd)
+{
+    return cmd.matches("q", "quit") || cmd.matches("qa", "qall");
+}
+
+bool FakeVimProxy::save()
+{
+    if (!hasChanges())
+        return true;
```
ouuan, line 394:
```suggestion
```

This is incorrect when the tab is untitled and the content is the same as the template.

You don't need to care about this, `MainWindow::save` checks everything.

Besides the mistake and duplicate code, it checks twice, which affects the performance.

ouuan (changes requested):

src/Settings/settings.json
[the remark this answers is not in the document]
ouuan: > For those who like to code in REPLACE mode will have to press `INS` to change cursor mode everytime new tabs are opened. This setting just helps them in it.

I'm wondering if this is helpful to anyone :thinking: Do you know anyone who likes to code in replace mode?

Anyhow, the name "Default cursor overwrites" really confused me.

ouuan:
We need more tests on Windows. I tested several commits and they all crashed on `:q`.

coder3101:
> We need more tests on Windows. I tested several commits and they all crashed on `:q`.

I'll investigate and fix it.

src/Settings/settings.json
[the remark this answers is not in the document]
coder3101: In vim mode there is a way to make cursor be in replace or insert mode. In all popular code editor you have ability to toggle Insert and Replace mode. 

As for the name, Do you think "Use replace mode for default cursor" is less confusing? If you want we can completely get rid of this setting as well. User will have to manually toggle it by pressing INS, the setting is nothing more than a convienience.

As for who uses it, I am sure people who code in old environment use it, like tourist.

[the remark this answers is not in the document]
ouuan: > In all popular code editor you have ability to toggle Insert and Replace mode.

Do you mean toggle during coding or toggle the default mode? Can you give an emample of a setting to toggle the default mode?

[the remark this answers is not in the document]
coder3101: > Do you mean toggle during coding or toggle the default mode? 

Toggling during coding.

> Can you give an emample of a setting to toggle the default mode?

I haven't found any because I haven't used Replace mode ever and bothered to look for such settings.

I will remove this setting.

coder3101:
I have fixed some warnings that were coming during compilation in Visual Studio and fixed the crash and removed and "Default cursor setting"

src/Core/FakeVimProxy.hpp
```diff
@@ -0,0 +1,103 @@
...
+  private:
+    static int firstNonSpace(QString const &);
+    void updateExtraSelections();
+    bool wantSaveAndQuit(FakeVim::Internal::ExCommand const &);
+    bool wantSave(FakeVim::Internal::ExCommand const &);
+    bool wantQuit(FakeVim::Internal::ExCommand const &);
+    bool save();
+    void cancel();
+    void setStatusBar();
+    void invalidate();
```
ouuan, line 86: Why the names are `cancel()` and `invalidate()`? I'd rather use `quit` and `forceQuit`, or combine them and make force a parameter.

coder3101: those names came from FakeVimExamples. I kept it that way, you can refactor them.

ouuan:
@coder3101 Please check this and other signals from `FakeVimHandler`:

https://github.com/cpeditor/FakeVim/blob/1dd90d7d61f990890bd1e6999bcdb8f0688e1c20/fakevim/fakevimhandler.cpp#L5614-L5630

ouuan:
I'm not sure if the recent upstream changes are useful or not.

src/appwindow.cpp
```diff
@@ -415,27 +430,30 @@ void AppWindow::openTab(const QString &path)
         }
     }
 
-    auto newWindow = new MainWindow(path, getNewUntitledIndex(), preferencesWindow, this);
```
coder3101, line 418: Instead of passing only `PreferenceWindow`, we pass Complete `AppWindow` and it has a getter `getPreferenceWindow()` which can be used to initialize the messagelogger.

ouuan: Doing changes here only creates more conflicts, mixing up pull requests, and making conversations hard to find.

coder3101:
> @coder3101 Please check this and [other signals](https://github.com/cpeditor/FakeVim/blob/1dd90d7d61f990890bd1e6999bcdb8f0688e1c20/fakevim/fakevimhandler.h#L159-L184) from `FakeVimHandler`:
> 
> https://github.com/cpeditor/FakeVim/blob/1dd90d7d61f990890bd1e6999bcdb8f0688e1c20/fakevim/fakevimhandler.cpp#L5614-L5630

Some of them are connected, some of them will be connected in future like `folds` and `completion` (after we have code folding and completion). Changing tab is implemented as custom command. Is there any signal that you think should be connected and handled in this PR?

coder3101:
The onces that I think could be done are:

1. Jump to local marks
2. Move to matching parenthesis (move by pressing % on one parenthesis)

Anything else?

ouuan:
> Changing tab is implemented as custom command.

Now `tabp` and `tabn` are valid commands with no effect.

And personally I think `tabp` and `tabn` are better than `prv` and `nxt`.

ouuan:
The cursor changes to `|` when the focus is lost. And I can't find how to get focus without changing the cursor position.

coder3101:
> The cursor changes to `|` when the focus is lost. And I can't find how to get focus without changing the cursor position.

How do you think we should handle it? I think when focus is lost the fakevimhandler becomes inactive and looses vim block cursor. There are two ways, a) we can make block cursor when focus is lost and restore it when it gets focus. So it does not becomes `|` when focus is lost. b) We automatically bring focus to editor when nothing else is in focus. Like after opening and closing the preference window the focus is lost from code editor and does not comes back to it, unless user clicks it.

ouuan:
> How do you think we should handle it? I think when focus is lost the fakevimhandler becomes inactive and looses vim block cursor. There are two ways, a) we can make block cursor when focus is lost and restore it when it gets focus. So it does not becomes `|` when focus is lost. b) We automatically bring focus to editor when nothing else is in focus. Like after opening and closing the preference window the focus is lost from code editor and does not comes back to it, unless user clicks it.

What do you think? I don't use vim very often, and I was simply thinking it looked strange. I'm not sure how it could be the most convenient.

ouuan:
1. The vimrc setting doesn't work as expected: if my vimrc setting is `set tabstop=4`, when opening a new tab the tabstop is 8, but if I run `set` manually (either with or without any arguments) the tabstop changes to 4.
2. QCodeEditor doesn't know tab width changes set by FakeVim, so `QCodeEditor::m_tabReplace` is not updated and several things including the column number are incorrect.

coder3101:
> 
> 
> I can't find how to get focus without changing the cursor position.

In my windows the focus is automatically set to the editor when mainwindow gets the focus. (i.e after preference window closes).
It could be a Qt bug :thinking:

[From docs](https://doc.qt.io/qt-5/focus.html#the-user-moves-the-focus-to-this-window) quoted below

> **The User Moves the Focus to This Window**
> In this situation the application must determine which widget within the window should receive the focus.
> This can be simple: If the focus has been in this window before, then the last widget to have focus should regain it. Qt does this automatically.
> If focus has never been in this window before and you know where focus should start out, call QWidget::setFocus() on the widget which should receive focus before you call QWidget::show() it. If you don't, Qt will pick a suitable widget.

ouuan:
> > I can't find how to get focus without changing the cursor position.
> 
> In my windows 10 the focus is automatically set to the editor when mainwindow gets the focus. (i.e after preference window closes).
> It could be a Qt bug on Linux or maybe KDE thinking .
> 
> [From docs](https://doc.qt.io/qt-5/focus.html#the-user-moves-the-focus-to-this-window) quoted below
> 
> > **The User Moves the Focus to This Window**
> > In this situation the application must determine which widget within the window should receive the focus.
> > This can be simple: If the focus has been in this window before, then the last widget to have focus should regain it. Qt does this automatically.
> > If focus has never been in this window before and you know where focus should start out, call QWidget::setFocus() on the widget which should receive focus before you call QWidget::show() it. If you don't, Qt will pick a suitable widget.

I don't think these are related to what I was talking about. I mean, when I click the code editor in order to get focus, the cursor position changes. If I'm using the real Vim and mouse wasn't on (i.e. I can't change cursor position by clicking), I can click the terminal without changing the cursor position.

It has nothing to do with the preferences window, I wan't talking about losing focus after opening and closing the preferences window, and in fact the focus is not lost after opening and closing the preferences window.

It's definetely not a bug. I haven't come up with any solution better than adding an option to disable using mouse to change cursor position. However, I'm not sure how to implement this yet.

ouuan:
How did you understand "I can't find how to get focus without changing the cursor position"?

coder3101:
I have fixed the cursor becoming `|` when lossing focus. The logic is to override QCodeEditor::focusOutEvent and making cursor block, when fakevim gets the focus. it sets the cursor back to its old state. ( | for insert block for normal)

coder3101:
> 
> 
> How did you understand "I can't find how to get focus without changing the cursor position"?

Do you mean mouse click shouln't change the cursor position? It's like disabling mouse click event for QCodeEditor.

ouuan:
> Do you mean mouse click shouln't change the cursor position? It's like disabling mouse click event for QCodeEditor.

"I haven't come up with any solution better than adding an option to disable using mouse to change cursor position."

I mean, it's good if I can use the mouse to change the cursor position while being able to use mouse to get focus to the code editor. However, I don't know if this is possible. Perhaps we can try locking the cursor position at the click which gets the focus.

coder3101:
> use mouse to get focus to the code editor without changing the cursor position. However, I don't know if this is possible.

Why not override `mousePressEvent()` and call `setFocus` and not propagate the `mousePressEvent()` to QTextEdit? This way, on mouse press we are only setting focus and nothing else.

ouuan:
> Why not override `mousePressEvent()` and call `setFocus` and not propagate the `mousePressEvent()` to QTextEdit? This way, on mouse press we are only setting focus and nothing else.

I've tried something like this and failed. `hasFocus()` in `mousePressEvent()` always returned `true`.

ouuan (changes requested):
https://github.com/cpeditor/cpeditor/pull/581#issuecomment-724609915

coder3101:
>  The vimrc setting doesn't work as expected: if my vimrc setting is `set tabstop=4`, when opening a new tab the tabstop is 8, but if I run `set` manually (either with or without any arguments) the tabstop changes to 4.

Yes, I acknowledge it. I'll work on it.

> QCodeEditor doesn't know tab width changes set by FakeVim, so QCodeEditor::m_tabReplace is not updated.

I think Tab key press is intercepted by Fakevim and it sets the correct width for it. Do you mean when fakevim is disabled (after it was enabled) the code's tab width changes?

> several things including the column number are incorrect.

Column numbers are incorrect? Could you please explain an example would be great. It is a consequence of one of above issue?

ouuan:
> Column numbers are incorrect? Could you please explain an example would be great.

The number of colums a tab takes is determined by `QCodeEditor::m_tabReplace`, so if the tab width is 4, no matter what the `tapstop` config of FakeVim is, the column number of `<tab><cursor>` is 5.

> is it a consequence of one of above issue?

I'm wondering why you splitted "2. QCodeEditor doesn't know tab width changes set by FakeVim, so `QCodeEditor::m_tabReplace` is not updated and several things including the column number are incorrect" into two parts. If they are not related, I would write it in `3.`.

coder3101:
I think the column numbers are computed in incorrect way.
Assume `*` as space character and `\t` as tab, `|` as cursor and tabstop as 4.
```
****| 
# Column = 5

\t|
# VSCode says Column = 5 but Vim/NeoVim says Column = 2
# Our QCodeEditor says column as 5, Fakevim tells column as 2
```
When Vim emulation is enabled, the Editor's Column number should use the Vim way to calculate column numbers.

ouuan:
> VSCode says Column = 5 but Vim/NeoVim says Column = 2

My vim provides two column numbers:

image

In the second one, a tab is how wide it displays.

ouuan:
A possible solution is to modify the code of FakeVim (`FakeVimHandler::Private::setTabSize`). We may need to include QCodeEditor in FakeVim.

coder3101:
> 
> 
> A possible solution is to modify the code of FakeVim (`FakeVimHandler::Private::setTabSize`). We may need to include QCodeEditor in FakeVim. Or FakeVim can emit a signal when the tab size changes.

third_party should not include any CP Editor thing. A better solution will be to hide our Row/Column indicator in vim mode and let fakevim show them in status bar. Currently, there are 2 row/col indicators in vim mode, one by fakevim in statusBar and another by us below the editor widget.

ouuan:
Well, it's my mistake. The column number is not related to QCodeEditor, it's calculated in `MainWindow::updateCursorInfo()` according to `SettingsHelper::getTabWidth()`.

coder3101:
> > The vimrc setting doesn't work as expected: if my vimrc setting is `set tabstop=4`, when opening a new tab the tabstop is 8, but if I run `set` manually (either with or without any arguments) the tabstop changes to 4.
> 
> Yes, I acknowledge it. I'll work on it.

I can't figure out why this does not work as expected. [See the example](https://github.com/cpeditor/FakeVim/blob/master/example/main.cpp) of Fakevim, the steps are mostly the same. `connectSignal -> initHandler -> loadvimrc`

ouuan:
I think we can support Tab Width, etc. in Vim emulation, while those settings can be overriden by .vimrc.

coder3101:
> I think we can support Tab Width, etc. in Vim emulation, while those settings can be overriden by .vimrc.

What else do you want to be enabled in addition to Tab Width?

coder3101:
This PR is ready for final review. I have fixed most of the issues you pointed. You can give it a try again and let me know if there are more changes required.

coder3101:
Old documentation PR has been closed and a new one has been opened here https://github.com/cpeditor/cpeditor.github.io/pull/115

coder3101:
Vim mode is for cpeditor 7.0 and has been ported to new code editor on #516

PR #598, opened 2020-09-19
chore: add --abbrev=8 for git describe

ouuan:
## Description

For an unknown reason, the default abbrev for CP Editor is 8 digits now. And somehow it's still 7 digits during CI ([with git 2.28.0](https://github.com/actions/virtual-environments/blame/757f01513062e82bbe8c5670c0b118fafe89e9bc/images/linux/Ubuntu1804-README.md#L183)), so [the CI failed](https://github.com/cpeditor/cpeditor/actions/runs/262574039). This should fix it.

It seems that the reason it is 7 digits during CI is that only the HEAD and its fathers are fetched, and it is 7 digits with only HEAD and its fathers, but 8 digits with all commits.

### DIFF
--- a/.ci/aur/git/PKGBUILD
+++ b/.ci/aur/git/PKGBUILD
@@ -37,7 +37,7 @@ md5sums=('SKIP' 'SKIP' 'SKIP' 'SKIP' 'SKIP' 'SKIP')
 
 pkgver() {
 	cd "$_pkgname"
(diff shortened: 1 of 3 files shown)


coder3101 (approved):

ouuan:
I found the reason: https://github.com/git/git/commit/e6c587c733b4634030b353f4024794b08bc86892

PR #606, opened 2020-09-22
fix(Settings): apply settings in the correct order

coder3101:
## Description
- Apply MainWindow settings after AppWindow Settings are applied.
- Delay applySettings in MainWindow.

## Related Issue
Fixes #604

## Motivation and Context
See #604

## How Has This Been Tested?
Arch Linux with Fusion Themes

## Type of changes

- [x] Bug fix (changes which fix an issue)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/cpeditor/cpeditor/blob/master/CONTRIBUTING.md) document.
- [x] I have tested these changes locally, and this fixes the bug/the new feature behaves as the expectation.
- [x] The settings file in the old version can be used in the new version after this change.
- [x] These changes only fix a single bug/introduces a single feature. (Otherwise, open multiple Pull Requests instead, unless these bugs/features are closely related.)
- [x] The commit messages are clear and detailed. (Otherwise, use `git reset` and commit again, or use `git rebase -i` and `git commit --amend` to modify the commit messages.)
- [x] These changes don't remove an existing feature. (Otherwise, add an option to disable this feature instead, unless it's necessary to remove this feature.)
- [x] If there are changes of the text displayed in the UI, I have wrapped them in `tr()` or `QCoreApplication::translate()`.
- [x] I have documented these changes in [CHANGELOG.md](https://github.com/cpeditor/cpeditor/blob/master/doc/CHANGELOG.md), or these changes are not notable.

PR #610, opened 2020-09-25
fix(cftool): Check for CF Tool

coder3101:
## Description
 Since #606 the settings are being applied later. Due to this, the variable named cftoolpath was empty when it was used to check the cf tool availability and a false error of no cf tool was being shown. With this PR we use SettingsHelper and let applySettings check and warn the availability of CF Tool

## Type of changes

## Checklist

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -277,15 +277,6 @@ void MainWindow::setCFToolUI()
             }
         });
     }
-    if (!Extensions::CFTool::check(SettingsHelper::getCFPath()))
-    {
-        qDebug() << "CF tool was not found";
-
-        submitToCodeforces->setEnabled(false);
-        log->error(tr("CF Tool"),
-                   tr("You will not be able to submit code to Codeforces because CF Tool is not installed or is "
-                      "not on SYSTEM PATH. You can set it manually in settings."));
-    }
 }
 
 int MainWindow::getUntitledIndex() const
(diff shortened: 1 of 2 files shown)


coder3101:
After #581, I will work on refactoring the MainWindow and AppWindow Structure. They are very annoying and confusing to me now.

ouuan:
Are any of these useful without #606?

coder3101:
Its improvement even without that PR. Since it gets rid of a member variable `cftoolpath`. Without #606, it should be titled as "Improve CF tool check"

coder3101:
category: refactor

ouuan:
We'll enable custom submit commands in the future, and I personally want to remove all checks on an executable before executing it.

About the codes, this PR actually adds one line, and I think `applySettings("Extensions/CF Tool", false);` in `setProblemURL` is weird.

coder3101:
> We'll enable custom submit commands in the future, and I personally want to remove all checks on an executable before executing it.

What how will you check the executable path then? Directly in Preferences before applying? If so, We can discard the PR

ouuan:
> What how will you check the executable path then?

When executing it.

ouuan:
I mean, we check the execution, instead of using another execution to check it.

ouuan:
Actually I was more referring the "diagonistics" in the mainwindow.

coder3101:
I get it.

coder3101:
> > What how will you check the executable path then?
> 
> When executing it.

Close In favour of above Idea

PR #612, opened 2020-09-26
fix(StyleManager): use palette instead of stylesheet for tooltips

### LINKED ISSUE #604: Message logger font automatically resets.
**Describe the bug**
After applying the message logger font, it stays applied only till cpeditor is open. When closing and reopening cpeditor, the message logger font shows an unknown font and unknown font size. Basically it resets my applied font to some other font on every launch. However in the settings it shows the correct name of the font which I had applied. And because of that I've to set the message logger font everytime.

**To Reproduce**
Set the UI style as auto/light/dark fusion and the message logger font to some distinguishable font. Then relaunch cpeditor and run a sample program.

**Screenshots**
After setting the message logger font-

Screenshot (73)

After closing and relaunching cpeditor-

Screenshot (74)

**Environment**

 - OS: Windows 10
 - CP Editor Version: 6.64/6.5 (Tried on both)

**Additional context**
None

ouuan:
## Description

Now the color of the tooltips are controlled by the palette, and the stylesheet is not changed.

## Related Issue

This fixes #604 and reverts #606.

## Motivation and Context

1. #604 is caused by unexpected behavior when setting the stylesheet. Instead of changing the order of applying settings, getting rid of the stylesheet has no side effects like #610.
2. I think the new palettes actually seem better than the old stylesheets.

## How Has This Been Tested?

On Arch Linux KDE and Windows, with Dark Fusion and Light Fusion and message logger font changed.

image

image

## Type of changes

- [x] Revert (revert previous changes)

## Checklist

### DIFF
--- a/src/appwindow.cpp
+++ b/src/appwindow.cpp
@@ -1010,6 +1010,12 @@ void AppWindow::onSettingsApplied(const QString &pagePath)
 {
     LOG_INFO("Apply settings for " << INFO_OF(pagePath));
 
+    for (int i = 0; i < ui->tabWidget->count(); ++i)
+    {
+        windowAt(i)->applySettings(pagePath, i == ui->tabWidget->currentIndex());
+        onEditorTextChanged(windowAt(i));
+    }
+
     auto pageChanged = [pagePath](const QString &page) { return pagePath.isEmpty() || pagePath == page; };
 
     if (pageChanged("Key Bindings"))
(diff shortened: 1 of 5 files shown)


ouuan:
In fact, I've been wondering about `darkPalette.setColor(QPalette::ToolTipBase, Qt::white);` and `lightPalette.setColor(QPalette::ToolTipBase, Qt::black);` for a long time.

src/appwindow.cpp
```diff
@@ -1010,6 +1010,12 @@ void AppWindow::onSettingsApplied(const QString &pagePath)
 {
     LOG_INFO("Apply settings for " << INFO_OF(pagePath));
 
+    for (int i = 0; i < ui->tabWidget->count(); ++i)
```
coder3101, line 1013: Why do you want to apply the settings to each tab then to whole application?

ouuan: Because there's no need to do the reverse, which could lead to unforeseen bugs.

coder3101: I changed the order because of some issues, I'll check if it happens after your palette based stylings and let you know.

coder3101: I checked its fine!

coder3101 (approved):
LGTM

PR #613, opened 2020-10-01
chore: move links from /doc to the website

ouuan:
## Description

Move links from `/doc` to the website.

Although the translations on the website are not finished, I think the documentation on the website is much better. We can start the translations on the website later.

I used Google translate and the old translations to translate some words into Russian.


coder3101 (approved):

coder3101:
I was just thinking that docs for some version is on url `cpeditor.org/v$major.minor` where is the docs for alpha or master branch? which version does cpeditor.org shows?

CMakeLists.txt
```diff
@@ -70,6 +70,8 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
       string(REGEX REPLACE "(.+)-(.+)-(.+)" "\\2" "COMMITS_SINCE_LAST_RELEASE" "${GIT_DESCRIBE}")
       if(NOT "${COMMITS_SINCE_LAST_RELEASE}" EQUAL "0")
         string(REGEX REPLACE "(.+)-(.+)-(.+)" "\\1.r\\2.\\3" "DISPLAY_VERSION" "${GIT_DESCRIBE}")
+      else()
+        string(REGEX REPLACE "([0-9]+\\.[0-9]+)\\..+" "v\\1" "MINOR_VERSION" "${GIT_DESCRIBE}")
```
coder3101, line 74: MINOR VERSION seems a bad name choice. It is confusing. Suppose a version is 6.7.2 the term MINOR VERSION of above version is just 7. While a correct name could be MAJOR_MINOR_VERSION which could be interpreted as v6.7 (includes both major and minor number).

src/appwindow.cpp
```diff
@@ -579,8 +579,7 @@ void AppWindow::on_actionSupportMe_triggered()
 
 void AppWindow::on_actionManual_triggered()
 {
-    QDesktopServices::openUrl(QUrl(
-        tr("https://github.com/cpeditor/cpeditor/blob/%1/doc/MANUAL.md").arg(GIT_COMMIT_HASH), QUrl::TolerantMode));
+    QDesktopServices::openUrl(QUrl(tr("https://cpeditor.org/%1/docs").arg(MINOR_VERSION), QUrl::TolerantMode));
```
coder3101, line 582: Can we guarantee that URL exists on alpha, beta, stable all?

ouuan:
Only releases have a non-empty `MINOR_VERSION`.

CMakeLists.txt
ouuan: I think "6" is a "major version number", "v6" is a "major version", "7" is a "minor version number", and "v6.7" is a "minor version".

ouuan: You can't say "7" in "6.7.2" is a "version".

coder3101: Fine then!

src/appwindow.cpp
coder3101: If minor version is empty (in case of alpha/aur) the URL is malformed. Maybe you can direct to default cpeditor.org in such cases

coder3101 (approved):

PR #615, opened 2020-10-01
chore: delete doc

ouuan:
## Description

Delete the doc directory.


coder3101:
Shouldn't the translated stuff be re-used in website before deleting them?

ouuan:
> Shouldn't the translated stuff be re-used in website before deleting them?

We are using git.

ouuan:
Do you mean we should show users with the old but translated version?

coder3101:
@ouuan Yes.

ouuan:
How? Where? I don't think it's a good idea to use the old version in the links for documentation.

ouuan:
I think we can either show users the English version, or wait until the translation is finished.

coder3101:
At least you can setup the Russian and Chinese version in docs on website and use these translation and then delete them here. For docs that are not yet translated you can keep the same English content and possibly add a warning "WIP" 

It will help other translators to just translate things instead of worrying about setup.

coder3101:
Okay

coder3101 (approved):

PR #624, opened 2020-10-06
refactor(Companion): Use QHttp server

### LINKED ISSUE #345: Competitive companion parser error
While parsing problems from competitve companion a JSON parser error reflects. It resolves on restart cp editor. But it is irritating.
Screenshot from 2020-05-13 15-39-32

 - OS: Ubuntu 18.04
 - CP Editor Version: 6.2.5

coder3101:
## Related Issue
Fixes #345 (Confirmation required)

## Motivation and Context
It is reported that in some cases the Competitive companion server is skipping some data or not at all accepting valid data on Chrome browser. This could not be reprodduced by us, but a possible solution was to use some other ready made Qt Http Server. In this PR, We use that server to listen for requests sent by Competitive Companion. This should solve the issue. (confirmation required)

## How Has This Been Tested?
Arch Linux with KDE

N/A
## Type of changes

## Checklist

- [x] I have documented these changes in [CHANGELOG.md](https://github.com/cpeditor/cpeditor/blob/master/CHANGELOG.md), or these changes are not notable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
--   We now use QHttp to listen for Competitive Companion data. This should fix an issue where in some cases CP Editor could not get full data from Google Chrome browsers. (#345 and #624)
+-   Fixed an issue where in some cases CP Editor could not get full data from Google Chrome browsers. (#345 and #624)
 
 ## v6.7
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## UNRELEASED
 
+### Added
+
+### Fixed
+
+-   We now use QHttp to listen for Competitive Companion data. This should fix an issue where in some cases CP Editor could not get full data from Google Chrome browsers. (#345 and #624)
+
 ## v6.7
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
--   Fixed an issue where in some cases CP Editor could not get full data from Google Chrome browsers. (#345 and #624)
+-   Fixed an issue that in some rare cases CP Editor could not get full data from Competitive Companion. (#345 and #624)
 
 ## v6.7
 
--- a/src/Extensions/CompanionServer.cpp
+++ b/src/Extensions/CompanionServer.cpp
@@ -55,44 +55,44 @@ bool CompanionServer::startListeningOn(int port)
     if (server)
         delete server;
     server = new qhttp::server::QHttpServer(this);
-    server->listen(
-        QString::number(port), [&, this](qhttp::server::QHttpRequest *req, qhttp::server::QHttpResponse *res) {
-            LOG_INFO("\n--> " << req->methodString() << " : %s" << qPrintable(req->url().toString().toUtf8()));
-            req->headers().forEach(
-                [](auto iter) { LOG_INFO(iter.key().constData() << " : " << iter.value().constData()); });
+    server->listen(QString::number(port),
+                   [&, this](qhttp::server::QHttpRequest *req, qhttp::server::QHttpResponse *res) {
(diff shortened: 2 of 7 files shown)


coder3101:
S*** Windows MSVC Compiler is failing. Will fix it when on Windows!

coder3101:
@IZOBRETATEL777 and @ouuan translations required.

src/Extensions/CompanionServer.cpp
```diff
@@ -18,148 +18,146 @@
...
+                           res->addHeader("connection", "close");
+                           res->addHeader("pragma", "no-cache");
+
+                           if (methodType != "POST")
+                           {
+                               SAFEWLOG("Companion", tr("Discarded %1 request").arg(methodType));
+                               res->setStatusCode(qhttp::ESTATUS_NOT_ACCEPTABLE);
+                               res->end();
+                               return;
+                           }
```
ouuan, line 78: What does this error mean? When does it happen? Why is it a warning?

ouuan:
Is it guaranteed that the `content-type` is not other words like `Content Type` on all platforms?

coder3101:
> Is it guaranteed that the `content-type` is not other words like `Content Type` on all platforms?

Yes as per QHttpServer documentation, the header is always lowercase. Moreover, the function `keyHasValue(a,b)` compares the key and value in case insensitive matter.

src/Extensions/CompanionServer.cpp
coder3101: When a GET/PUT/DELETE or any other non POST request is made to the server port. We Gracefully ignore it and issue a warning. We could have ignored it all together even without logging a warning but if this happens user should know that something else in their computer is making some request to the Port which is suppose to listen for only Competitive Companion POST data.

coder3101: One way to trigger this warning is that you could visit `localhost:10045` (port of competitive companion) from your web browser. It sends a GET request and CP Editor will issue the warning. From development point of view it is another way to check if server is actually listening on the port.

src/Extensions/CompanionServer.cpp
```diff
@@ -18,148 +18,143 @@
...
+    server->listen(QString::number(port), [this](qhttp::server::QHttpRequest *req, qhttp::server::QHttpResponse *res) {
+        LOG_INFO("\n--> " << req->methodString() << " : " << qPrintable(req->url().toString().toUtf8()));
+        req->headers().forEach(
+            [](auto iter) { LOG_INFO(iter.key().constData() << " : " << iter.value().constData()); });
+
+        const QString methodType = req->methodString();
+        const bool isJson = req->headers().keyHasValue("content-type", "application/json");
+        req->collectData();
+
+        req->onEnd([=] {
```
coder3101, line 67: Why not capture by reference?

coder3101:
@IZOBRETATEL777 or @sadykhzadeh Could you please add Russian translation? 

A translation of these two statements is required.

1. "The request received is not JSON"
2. "A %1 request is received and ignored"

If it is not possible for you to sync this PR. You could atleast reply to this thread with the translations for the above statement.

src/Extensions/CompanionServer.cpp
ouuan: I think they are local variables and will be destroyed at the end of `CompanionServer::startListeningOn`, which results in undefined behavior.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ### Fixed
 
--   We now use QHttp to listen for Competitive Companion data. This should fix an issue where in some cases CP Editor could not get full data from Google Chrome browsers. (#345 and #624)
+-   Fixed an issue where in some cases CP Editor could not get full data from Google Chrome browsers. (#345 and #624)
```
ouuan, line 7:
```suggestion
-   Fixed an issue that in some cases CP Editor could not get full data from Competitive Companion. (#345 and #624)
```

We have no evidence that it only happens in Chrome.

ouuan (approved):

PR #627, opened 2020-10-10
refactor(Settings): move old to JSON and rename Test Case Length Limit

ouuan:
## Description

1. The old settings keys are moved from SettingsUpdater to settings.json. It reduces the modification when changing the name of a setting, especially when a setting has multiple old names.
2. Rename "Load Test Case Length Limit" to "Display Test Case Length Limit", which is a better name since #491.

### DIFF
--- a/src/Settings/SettingsUpdater.cpp
+++ b/src/Settings/SettingsUpdater.cpp
@@ -19,63 +19,41 @@
 #include "Core/SessionManager.hpp"
 #include "Settings/SettingsManager.hpp"
 #include "generated/SettingsHelper.hpp"
+#include <QDebug>
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
-#include <QMap>
 #include <QSettings>
-
-const static QMap<QString, QString> updateInfo = {
-    {"tab_stop", "Tab Width"},
-    {"font", "Editor Font"},
-    {"clang_format_binary", "Clang Format/Path"},
-    {"clang_format_style", "Clang Format/Style"},
-    {"template_cpp", "C++/Template Path"},
-    {"compile_cpp", "C++/Compile Command"},
-    {"runtime_cpp", "C++/Run Arguments"},
-    {"template_java", "Java/Template Path"},
-    {"compile_java", "Java/Compile Command"},
-    {"runtime_java", "Java/Run Arguments"},
-    {"run_java", "Java/Run Command"},
-    {"template_java", "Python/Template Path"},
-    {"runtime_python", "Python/Run Arguments"},
-    {"run_python", "Python/Run Command"},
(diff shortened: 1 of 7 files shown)


src/Settings/SettingsUpdater.cpp
```diff
@@ -19,63 +19,41 @@
...
-
 void SettingsUpdater::updateSetting(QSettings &setting)
 {
-    for (const auto &old : updateInfo.keys())
+#ifdef QT_DEBUG
+    // Check for key conflicts
+    QSet<QString> keys;
+    for (const auto &si : qAsConst(SettingsInfo::settings))
+    {
+        auto addKey = [&](const QString &key) {
```
coder3101, line 35: Why create a new lambda for each key `si`? Can't you create just one and pass arguments to it for different?

translations/ru_RU.ts
```diff
@@ -2523,6 +2509,15 @@ The program will be killed if it doesn&apos;t terminate in the time limit.</sour
...
         <translation>Показывать только моноширинные шрифты при выборе</translation>
     </message>
+    <message>
+        <source>Display Test Case Length Limit</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>The maximum number of characters in a test case to be displayed.
+A test case will be elided and read-only if it&apos;s too long.</source>
+        <translation type="unfinished"></translation>
```
coder3101, line 2519: @IZOBRETATEL777 translation needed here.

src/Settings/settings.json
```diff
@@ -657,11 +767,15 @@
         "tip": "The maximum number of characters in a source file to open.\nA source file won't be opened if it's too long."
     },
     {
-        "name": "Load Test Case Length Limit",
+        "name": "Display Test Case Length Limit",
```
coder3101, line 770: Although very minor change, I think it should be reflected in the `cpeditor/docs` as well.

ouuan: It won't be in the changelog, but the docs for this setting will be updated.

ouuan: I noticed the name of this setting when reviewing cpeditor/cpeditor.github.io#76

coder3101: > It won't be in the changelog, but the docs for this setting will be updated.

Yes, I mean the doc in the website for English and Chinese (possibly) could be updated to "Display" instead of "Load".

This is just another way to make sure that we keep website doc up to date as we merge PRs.

coder3101:
"awaiting for docs" label will be removed when you open a PR that updates the doc. 

We should make sure that no PR get merged with the above label. It will mean CP Editor has some feature that english doc does not mentions.

IZOBRETATEL777 (approved):

coder3101 (approved):

PR #632, opened 2020-10-17
refactor(Checker): use TResult in testlib.h instead of a list of numbers

ouuan:
## Description

Use TResult in `testlib.h` instead of a list of numbers.

## Motivation and Context

Use `-Wswitch` to generate a compilation warning when there are unhandled items in TResult.

## How Has This Been Tested?

On Arch Linux with testlib.h checkers and a fake checker which returns an unknown value.

## Checklist

### DIFF
--- a/src/Core/Checker.cpp
+++ b/src/Core/Checker.cpp
@@ -24,6 +24,7 @@
 #include <QFile>
 #include <QTemporaryDir>
 #include <generated/SettingsHelper.hpp>
+#include <testlib.h>
 
 namespace Core
 {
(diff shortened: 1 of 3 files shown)


src/Core/Checker.cpp
```diff
@@ -170,29 +171,36 @@ void Checker::onRunFinished(int index, const QString &, const QString &err, int
...
-        // This exit code is a normal exit code of a testlib checker, means WA or something else
+        return;
+
+    case _wa:
+    case _pe:
+    case _fail:
+    case _dirt:
+    case _points:
+    case _unexpected_eof:
+    case _partially:
```
coder3101, line 188: It would be great if for each special WA, the checker could report some reason. Something like `Checker[3]: Wrong answer, Unexpected End of file, output has less contents than expected`.

Nothing special should be displayed if WA is because of output/expected mismatch.

ouuan: That's in the stderr of the checker.

coder3101: If checker returns `_unexpected_eof` we can simply `emit(index, WA)` we could `log->info("Checker [%1]".arg(index), "Early EOF encountered")` before we emit WA.

In this way, the user has some more information that WA because of Early EOF.

ouuan: As I said, the more detailed information is in the stderr of the checker.

coder3101: Is stderr of checker logged in the Messagelogger?

ouuan: > Is stderr of checker logged in the Messagelogger?

Yes. You'd better read the codes or have a try. I said twice it's in the stderr, which means it's in the message logger.

coder3101 (approved):

PR #659, opened 2020-11-06
feat: link to settings in the message, remove diagnostics, improve message

ouuan:
## Description

1. Link to the settings in the message logger, in the messages containing the path of the setting.
2. Remove diagnostics (which is not very related, but I decided to do this when trying to modify the messages of it).
3. Improve some messages.
4. Some other refactors.

## Motivation and Context

Make it easy to find the settings you need to change, especially for new users.

## How Has This Been Tested?

On Arch Linux and Windows 10.

image

## Checklist

## Additional text

@IZOBRETATEL777 please update the translations.

### DIFF
--- a/src/Core/MessageLogger.cpp
+++ b/src/Core/MessageLogger.cpp
@@ -17,16 +17,24 @@
 
 #include "Core/MessageLogger.hpp"
 #include "Core/EventLogger.hpp"
+#include "Settings/PreferencesWindow.hpp"
 #include "generated/SettingsHelper.hpp"
 #include <QDateTime>
 
-MessageLogger::MessageLogger(QWidget *parent) : QTextBrowser(parent)
+MessageLogger::MessageLogger(PreferencesWindow *preferences, QWidget *parent)
+    : QTextBrowser(parent), preferencesWindow(preferences)
 {
+    connect(this, SIGNAL(anchorClicked(const QUrl &)), this, SLOT(onAnchorClicked(const QUrl &)));
     setOpenExternalLinks(true);
 }
 
 void MessageLogger::message(const QString &head, const QString &body, const QString &color, bool htmlEscaped)
 {
+    if (htmlEscaped && body.contains("<a href="))
+    {
+        LOG_WARN("The message contains \"<a href=\" but htmlEscaped is enabled. " << INFO_OF(body));
+    }
+
     QString newHead, newBody;
     if (htmlEscaped)
     {
@@ -76,4 +84,12 @@ void MessageLogger::error(const QString &head, const QString &body, bool htmlEsc
 {
     LOG_INFO(INFO_OF(head) << INFO_OF(body));
     message(head, body, "red", htmlEscaped);
+}
+
+void MessageLogger::onAnchorClicked(const QUrl &link)
+{
+    auto url = link.toString();
+    LOG_INFO(INFO_OF(url));
+    if (url.startsWith("#Preferences/"))
+        preferencesWindow->open(url.mid(13));
 }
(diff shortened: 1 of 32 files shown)


src/Core/MessageLogger.cpp
```diff
@@ -17,23 +17,25 @@
...
+    setOpenExternalLinks(true);
 }
 
 void MessageLogger::message(const QString &head, const QString &body, const QString &color, bool htmlEscaped)
 {
+    if (body.contains("<a href='#Preferences/"))
+    {
+        LOG_WARN("The message contains \"<a href='#Preferences/\", htmlEscaped is automatically disabled.");
+        htmlEscaped = false;
+    }
```
coder3101, line 37: Why not htmlEscape everything except `<a href...>` (anchor) It will be much safer than keeping whole message unescaped. 

So,

```
Please do this here | <a href=..> | or do this.
--------escaped-----| ----html----| --escaped--
```

This way, we can open external links to in browser and other stuff like docs.

src/Core/MessageLogger.cpp
```diff
@@ -64,29 +66,31 @@ void MessageLogger::message(const QString &head, const QString &body, const QStr
...
 
 void MessageLogger::error(const QString &head, const QString &body, bool htmlEscaped)
 {
-    LOG_INFO("MessageLogger Error " << INFO_OF(head) << INFO_OF(body));
+    LOG_INFO(INFO_OF(head) << INFO_OF(body));
     message(head, body, "red", htmlEscaped);
 }
 
-void MessageLogger::clear()
+void MessageLogger::onAnchorClicked(const QUrl &link)
```
coder3101, line 90: Make sure that you can open external links like to website or other URLs. I am not sure but it seems this slot does nothing for normal URLs, only opens preference page links.

coder3101 (changes requested):

src/Core/MessageLogger.cpp
ouuan: I don't think it's very easy to parse a `<a/>` element.

Both htmlEscaped and this auto-unescape are designed for compilation errors, where the messages may contain `#include <considered as an HTML tag>` and the messages are sent in signals and logged in MainWindow, etc. 

So we can let Compiler log messages instead of sending signals, explicitly set htmlEscaped to false and remove this auto-unescape.

ouuan: The messages in the MainWindow and the Checker are different, so I decided to use a different signal for messages from the compiler, and only escape the messages from the compiler.

ouuan:
We can merge this and other PRs first, and @IZOBRETATEL777 can update the translations later.

coder3101:
> We can merge this and other PRs first, and @IZOBRETATEL777 can update the translations later.

We should open an issue and assign it to him.

ouuan:
> We should open an issue and assign it to him.

I don't think it's necessary. Unlike the website, Qt Linguist keeps track of unfinished translations, so we don't need to track it in the issues.

coder3101:
> > We should open an issue and assign it to him.
> 
> I don't think it's necessary. Unlike the website, Qt Linguist keeps track of unfinished translations, so we don't need to track it in the issues.

Having an issue can attract other contributors who are willing to translate the application.

ouuan:
> Having an issue can attract other contributors who are willing to translate the application.

We may have a pinned issue or something which is never closed.

coder3101:
> > Having an issue can attract other contributors who are willing to translate the application.
> 
> We may have a pinned issue or something which is never closed.

I'll open the issue now and you can pin and edit the text if you want.

ouuan:
Or we can simply mention it in CONTRIBUTING.md.

coder3101 (approved):

PR #662, opened 2020-11-09
refactor: validate test case indices

ouuan:
## Description

Validate test case indices.

## Motivation and Context

Make it safer.

## How Has This Been Tested?

On Arch Linux.

## Checklist

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -201,6 +201,12 @@ void MainWindow::run()
 
 void MainWindow::run(int index)
 {
+    if (index < 0 || index >= testcases->count())
+    {
+        LOG_WARN(INFO_OF(index) << INFO_OF(testcases->count()));
+        return;
+    }
+
     auto tmp = new Core::Runner(index);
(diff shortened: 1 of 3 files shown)


src/mainwindow.cpp
```diff
@@ -201,6 +201,12 @@ void MainWindow::run()
 
 void MainWindow::run(int index)
 {
+    if (index < 0 || index >= testcases->count())
+    {
+        LOG_WARN(INFO_OF(index) << INFO_OF(testcases->count()));
+        return;
+    }
+
```
coder3101, line 209: When can indices be out of range?

ouuan: When we make mistakes. It shouldn't be out of range.

ouuan:
We'd better use `LOG_DEV` from #663 :thinking:

src/mainwindow.cpp
coder3101: Okay, so it's for finding issues quickly.

coder3101 (approved):

coder3101 (approved):

PR #688, opened 2020-11-22
feat: copy output to expected and copy test cases

ouuan:
## Description

1. Some refactors of the words.
2. The `Singleton` Util class. C++ 17 is also enabled for `is_base_of_v`. We can easily get rid of it, but why not use C++ 17?
3. Copy output to expected.
4. Copy checked testcases.

## Motivation and Context

1. "Copy output to expected" is useful when you want to use the output of one code as the expected for another.
2. "Duplicate Tab" can copy an existing tab to a new tab, but it can't copy test cases to an existing tab.

## How Has This Been Tested?

On Arch Linux.

## Checklist

## Additional text

@IZOBRETATEL777 translations need update


src/Util/Singleton.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <type_traits>
+
+namespace Util
+{
+
+template <typename T, typename D = T> class Singleton
+{
+    friend D;
+    static_assert(std::is_base_of_v<T, D>, "T should be a base type for D");
```
coder3101, line 31:
```suggestion
    static_assert(std::is_base_of<T, D>::value, "T should be a base type for D");
```
This works upto C++11

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ set(CMAKE_AUTOUIC ON)
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTORCC ON)
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
coder3101, line 15:
```suggestion
set(CMAKE_CXX_STANDARD 14)
```

Let's not update it unless absolutely necessary.

src/Util/Singleton.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * (at your option) any later version.
+ *
+ * I will not be responsible if CP Editor behaves in unexpected way and
+ * causes your ratings to go down and or lose any important contest.
+ *
+ * Believe Software is "Software" and it isn't immune to bugs.
+ *
+ */
+
+// This class is from https://stackoverflow.com/a/46178158
```
coder3101, line 18: You don't really need to mention it since it's a general Curiously Recurring template pattern for static polymorphism. If you have copy pasted the code, you can keep it.

src/Util/Singleton.hpp
```diff
@@ -0,0 +1,49 @@
...
+#define SINGLETON_HPP
+
+#include <type_traits>
+
+namespace Util
+{
+
+template <typename T, typename D = T> class Singleton
+{
+    friend D;
```
coder3101, line 30: Why do you want `D` to be friend class?

src/Util/Singleton.hpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        static D inst;
+        return inst;
+    }
+
+  private:
+    Singleton() = default;
+    ~Singleton() = default;
+    Singleton(const Singleton &) = delete;
+    Singleton &operator=(const Singleton &) = delete;
```
coder3101, line 44: You can also delete move assignment and move construction. I think it is implicitly deleted but being explicit is no harm.

coder3101 (changes requested):
Everything else looks fine.

src/Util/Singleton.hpp
ouuan: Why not use C++17?

ouuan:
Well, I have no time now. I don't think any of your suggestions is necessary, nor any of them is bad. You can commit them if you want, I won't.

CMakeLists.txt
coder3101: Why not use C++17?

I do support using new language features and take full advantage of it. But there must be a solid reason for upgrade like let's say `std::filesystem` (in C++17), or if something is very hard to implement in older version of c++ like iterator over compile time list. Or most important factor could be compile time stuff with constexpr, but it has no effect in GUI programming.

I do not think `_v` vs `::value` is enough of a reason for upgrade now.

coder3101: In addition to all these, there isn't much of language feature changes since C++11 to C++17. Most of them were library changes, next major update C++20, will have enough reason to upgrade. modules, co-routines, ranges are big updates.

coder3101:
> Well, I have no time now. I don't think any of your suggestions is necessary, nor any of them is bad. You can commit them if you want, I won't.

Sure, I'll do and merge them when I get time.

src/Util/Singleton.hpp
coder3101, line 18:
```suggestion

```

src/Util/Singleton.hpp
coder3101, line 30:
```suggestion

```

src/Util/Singleton.hpp
coder3101, line 44:
```suggestion
    Singleton &operator=(const Singleton &) = delete;
    Singleton &operator=(Singleton &&) = delete;
```

src/Util/Singleton.hpp
```diff
@@ -0,0 +1,49 @@
...
+    static T &instance()
+    {
+        static D inst;
+        return inst;
+    }
+
+  private:
+    Singleton() = default;
+    ~Singleton() = default;
+    Singleton(const Singleton &) = delete;
```
coder3101, line 43:
```suggestion
    Singleton(const Singleton &) = delete;
    Singleton(Singleton &&) = delete;    
```

coder3101 (approved):

PR #693, opened 2020-11-24
refactor: Use clang-tidy to check code quality

coder3101:
##  Description
Added clang-tidy checks along with compilation if `USE_CLANG_TIDY` variable is defined.(default it is off).
It slows the process for compilation. All list of checks could be found inside CMakeLists.txt

N/A

## Motivation and Context
Codacy fails frequently, it is not even close to clang-tidy in terms of code quality checks.

## How Has This Been Tested?
Arch KDE

N/A

## Checklist

## Additional text
N/A

## POST MERGE STEP
**Remove Codeacy App from the repository.**


coder3101:
Seems like macOS and Linux machine to not have clang-tidy installed. Here are some checklist.
- [ ]  Install clang-tidy on Linux and macOS
- [ ]  Fix issues reported by clang-tidy on windows
- [ ]  Modify clang-tidy rules, so that CI fails if clang-tidy reports anything quality issues.

coder3101:
@ouuan To help you ease review, I will comment on some type of refactoring. This is a big refactor so it should help you understand the motivation for each type of change.

CMakeLists.txt
```diff
@@ -102,6 +107,14 @@ file(COPY ${CMAKE_SOURCE_DIR}/translations/translations.qrc DESTINATION ${CMAKE_
 
 set_property(SOURCE ${CMAKE_BINARY_DIR}/generated/SettingsHelper.hpp ${CMAKE_BINARY_DIR}/generated/SettingsInfo.cpp PROPERTY SKIP_AUTOGEN ON)
 
+# Perform clang-tidy checks after we have build all the third_party dependencies.
+if(USE_CLANG_TIDY)
+	message(STATUS "Clang tidy checks will be performed")
+	
+	set(CMAKE_CXX_CLANG_TIDY 
+		"clang-tidy;-header-filter=^(?!.*\\bthird_party\\b).*$;-checks=-*,bugprone-*,cppcoreguidelines-*,modernize-*,performance-*,readability-*,-modernize-use-trailing-return-type,-bugprone-suspicious-include,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-avoid-magic-numbers,-readability-magic-numbers,-cppcoreguidelines-owning-memory,-readability-braces-around-statements,-cppcoreguidelines-macro-usage,-cppcoreguidelines-avoid-c-arrays,-modernize-avoid-c-arrays,-readability-implicit-bool-conversion,-readability-implicit-bool-conversion,-cppcoreguidelines-pro-type-static-cast-downcast,-cppcoreguidelines-special-member-functions,-cppcoreguidelines-avoid-non-const-global-variables;-warnings-as-errors=*")
```
coder3101, line 115: Every check that starts with `-name-of-check` is disabled.  Every check that starts with `name-of-checks-*` enables all checks with those category. Different checks are separated with `,`.

We treat all warning as error so that compilation fails when clangd tidy reports something.

We exclude every header that is inside `third_party`

src/Core/Checker.cpp
```diff
@@ -21,10 +21,10 @@
 #include "Core/MessageLogger.hpp"
 #include "Core/Runner.hpp"
 #include "Util/FileUtil.hpp"
+#include "third_party/testlib/testlib.h"
```
coder3101, line 24: Full name to library makes sure that clang-tidy knows this header is third_party and it should not check for issues from this header. It is due to this that all third_party headers now have to start from `third_party/dep/header.hpp` instead of just `header.hpp`.

src/Core/Checker.cpp
```diff
@@ -44,12 +44,10 @@ Checker::Checker(const QString &path, MessageLogger *logger, QObject *parent) :
 
 Checker::~Checker()
 {
-    if (compiler)
-        delete compiler;
+    delete compiler;
```
coder3101, line 47: Deleting nullptr has no effect as per standard c++, so check is not required.

src/Core/Checker.cpp
```diff
@@ -182,7 +179,8 @@ void Checker::onCompilationKilled()
     disconnect(compiler, &Compiler::compilationErrorOccurred, this, &Checker::onCompilationErrorOccurred);
 }
 
-void Checker::onRunFinished(int index, const QString &, const QString &err, int exitCode, int, bool tle)
+void Checker::onRunFinished(int index, const QString & /*unused*/, const QString &err, int exitCode, int /*unused*/,
```
coder3101, line 182: Every argument in a function definition should be named or marked /*unused*/ to signify this argument isn't used.

src/Core/Checker.cpp
```diff
@@ -315,7 +313,7 @@ void Checker::check(int index, const QString &input, const QString &output, cons
             Util::saveFile(expectedPath, expected, tr("Checker"), false, log))
         {
             // if files are successfully saved, run the checker
-            auto tmp = new Runner(index);
+            auto *tmp = new Runner(index);
```
coder3101, line 316: Qualified `auto`, so that `decltyp(auto)` returns only type and not pointer to type. It is modern convention.

src/Core/Checker.cpp
```diff
@@ -329,7 +327,7 @@ void Checker::check(int index, const QString &input, const QString &output, cons
     }
 }
 
-QString Checker::head(int index) const
+QString Checker::head(int index)
```
coder3101, line 330: It can be made static because this function doesn't uses any member of this class

src/Core/EventLogger.cpp
```diff
@@ -33,11 +33,11 @@ namespace Core
 QFile Log::logFile;
 QTextStream Log::logStream;
 
-int Log::NUMBER_OF_LOGS_TO_KEEP = 50;
-int Log::MAXIMUM_FUNCTION_NAME_SIZE = 30;
-int Log::MAXIMUM_FILE_NAME_SIZE = 30;
-QString Log::LOG_DIR_NAME = "cpeditorLogFiles";
-QString Log::LOG_FILE_NAME = "cpeditor";
+const int Log::NUMBER_OF_LOGS_TO_KEEP = 50;
```
coder3101, line 36: Global variables should be const whenever possible.

src/Core/EventLogger.hpp
```diff
@@ -87,11 +87,11 @@ class Log
...
     /**
      * @brief initialize the event logger
-     * @param instanceCount the instance ID provided by SingleApplication, to distinct processes from each other
-     * @param dumpToStderr whether to print the logs into stderr or not
+     * @param instance the instance ID provided by SingleApplication, to distinct processes from each other
+     * @param dumptoStderr whether to print the logs into stderr or not
      * @note this should be called only once before logging anything
      */
-    static void init(int instanceCount, bool dumpToStderr = false);
+    static void init(int instance, bool dumptoStderr = false);
```
coder3101, line 94: The name of argument in definition and declaration should be exactly same.

src/Core/MessageLogger.cpp
```diff
@@ -33,7 +33,8 @@ void MessageLogger::message(const QString &head, const QString &body, const QStr
     LOG_WARN_IF(body.contains("<a href") && htmlEscaped,
                 "The message contains \"<a href\", but htmlEscaped is enabled.");
 
-    QString newHead, newBody;
+    QString newHead;
```
coder3101, line 36: Multiple declaration in one line reduces readability

src/Core/Runner.cpp
```diff
@@ -115,7 +113,7 @@ void Runner::runDetached(const QString &tmpFilePath, const QString &sourceFilePa
     // use apple script on Mac OS
     runProcess->setProgram("osascript");
     runProcess->setArguments({"-l", "AppleScript"});
-    QString script = "tell app \"Terminal\" to do script \"" +
+    QString script = R"(tell app "Terminal" to do script ")" +
```
coder3101, line 116: Use raw string literals instead of escape.

src/Core/SessionManager.cpp
```diff
@@ -87,7 +87,8 @@ void SessionManager::restoreSession(const QString &path)
     auto oldSize = app->size();
     app->setUpdatesEnabled(false);
 
-    for (auto const &tab : tabs)
+    for (auto const &tab : tabs) // NOLINT: loop variable 'tab' is always a copy because the range of type 'QJsonArray'
```
coder3101, line 90: To suppress a warning put in that line as comment `NOLINT: <you can put here reason as to why you supressed this report instead of fixing it`.

src/Core/SessionManager.cpp
```diff
@@ -113,7 +114,7 @@ void SessionManager::setAutoUpdateSession(bool shouldAutoUpdate)
         timer->stop();
 }
 
-void SessionManager::setAutoUpdateDuration(unsigned int duration)
+void SessionManager::setAutoUpdateDuration(int duration)
```
coder3101, line 117: setInterval takes int as argument.

src/Extensions/CompanionServer.cpp
```diff
@@ -126,7 +124,7 @@ void CompanionServer::updatePort(int port)
 
 void CompanionServer::parseAndEmit(QByteArray &data)
 {
-    QJsonParseError error;
+    QJsonParseError error{};
```
coder3101, line 127: Must be initialised. Default initialisation

src/Extensions/LanguageServer.cpp
```diff
@@ -217,7 +216,7 @@ QCodeEditor::SeverityLevel LanguageServer::lspSeverity(int in)
     return QCodeEditor::SeverityLevel::Error;
 }
 
-void LanguageServer::initializeLSP(QString filePath)
+void LanguageServer::initializeLSP(QString const &filePath)
```
coder3101, line 219: Use const reference whenever possible.

src/Settings/CodeSnippetsPage.hpp
```diff
@@ -34,7 +34,7 @@ class CodeSnippetsPage : public PreferencesPage
     Q_OBJECT
 
   public:
-    explicit CodeSnippetsPage(const QString &language, QWidget *parent = nullptr);
+    explicit CodeSnippetsPage(QString language, QWidget *parent = nullptr);
```
coder3101, line 37: Modernization, pass by value and then initialise with `std::move()`

src/Settings/DefaultPathManager.cpp
```diff
@@ -26,8 +26,7 @@ QMap<QString, QString> DefaultPathManager::defaultPath;
 
 QString DefaultPathManager::defaultPathForAction(const QString &action)
 {
-    const auto result =
-        convertPath(SettingsManager::get(QString("Default Path/Action/%1/Uses").arg(action)).toString());
+    auto result = convertPath(SettingsManager::get(QString("Default Path/Action/%1/Uses").arg(action)).toString());
```
coder3101, line 29: returning const result prevent compiler from optimising return value.

src/Settings/PreferencesWindow.cpp
```diff
@@ -500,11 +501,12 @@ QTreeWidgetItem *PreferencesWindow::getTopLevelItem(const QString &text) const
     return nullptr;
 }
 
-QTreeWidgetItem *PreferencesWindow::getChild(QTreeWidgetItem *item, const QString &text) const
+QTreeWidgetItem *PreferencesWindow::getChild(QTreeWidgetItem *item, // NOLINT: It can be static
```
coder3101, line 504: This method can be made static, but since I have no idea about these codes, it is upto you to make them static and remove that `//NOLINT` suppress.

src/Settings/SettingsInfo.hpp
```diff
@@ -65,7 +65,7 @@ class SettingsInfo
   private:
     static QList<SettingInfo> settings;
 
-    friend class SettingsUpdater;
+    friend struct SettingsUpdater;
```
coder3101, line 68: `SettingsUpdater` is defined as `struct` so friend declaration should be `struct` not `class`

src/Settings/SettingsManager.cpp
```diff
@@ -181,32 +169,30 @@ void SettingsManager::saveSettings(const QString &path)
...
-    else if (def->contains(key))
+    if (def->contains(key))
         return def->value(key);
-    else
-    {
-        if (!noUnknownKeyWarning.contains(key))
-            LOG_DEV("getting unknown key: " << key);
-        return QVariant();
-    }
+
```
coder3101, line 178:
```cpp
if(condition)
    return something;
else 
    return otherthing;
```

`else` is redundant and should be removed.

```cpp
if(condition)
    return something;
return other_thing;
```

src/Settings/StringListsItem.cpp
```diff
@@ -114,14 +114,14 @@ StringListsItem::StringListsItem(const QVariantList &cols, QWidget *parent) : QW
         int cols = table->columnCount();
         int rows = table->rowCount();
         auto items = table->selectedItems();
-        if (items.size())
+        if (!items.empty())
```
coder3101, line 117: Use empty to check for size, it improves code readability.

src/Settings/ValueWrapper.cpp
```diff
@@ -108,7 +100,7 @@ void PlainTextEditWrapper::set(QString s)
 
 void ComboBoxWrapper::init(QWidget *parent, QVariant param)
 {
-    QComboBox *item = new QComboBox(parent);
+    auto *item = new QComboBox(parent);
```
coder3101, line 103: use auto whenever possible to avoid naming types multiple times

coder3101:
This PR should be merged today, because if we develop on other branches, this will inevitably cause a cascade of merge conflicts.

ouuan:
Can we run it separately instead of during building?

coder3101:
> Can we run it separately instead of during building?

Yes, we can export `compile_commands.json` by cmake and then use it to run clang tidy. Problem is that `compile_command.json` will also include third_party sources, which will also be checked, which we do not want.

ouuan:
I don't like CI tests of 6~10 minutes. It will be much better if it can be sped up.

coder3101:
Do you think running it separately will speed up?

coder3101:
You should review and check the code changes, no matter how we run clang tidy the analysis will still be same.

coder3101:
Now you can create a action that runs `tools/clangTidyCheck` and it will check for issues inside all our sources. It takes about 4 min without any parallelism.

src/Settings/DefaultPathManager.cpp
ouuan: I think `const` improves readability :thinking: And Qt has Implicit Sharing.

tools/clangTidyCheck.sh
```diff
@@ -0,0 +1,19 @@
...
+	echo "clang-tidy not installed!";
+	exit 1;
+fi
+
+# Update compile_commands.json file
+mkdir -p build
+cd build
+cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=1
+mv compile_commands.json ..
+cd ..
```
coder3101, line 17: Running `cmake ..` requires that Qt is installed in workflow, if you want to skip this step. We can put actual compile_commands.json inside our source root. We can use linux CI to check for up to date `compile_commands.json`, just like we check for up to date translations.

A drawback would be that clang-tidy checks could only be performed in linux so any block of code with OS specifier like windows, macOS will not be performed.

src/Settings/DefaultPathManager.cpp
coder3101: The reason why this comes out as performance issue because usually compiler tries to move the return value to caller so no copy is made for returned value. If you make return variable `const` the compiler cannot perform move and is forced to return the value by copy. It is called return value optimization.

coder3101:
> Now you can create a action that runs `tools/clangTidyCheck` and it will check for issues inside all our sources. It takes about 4 min without any parallelism.

It has one more issue, when you run the file you need to make sure that all headers are present and could be found by clang-tidy, it means generated/*.cpp/hpp should also be available, but they are generated when doing actual build.

src/Settings/DefaultPathManager.cpp
ouuan: > The reason why this comes out as performance issue because usually compiler tries to move the return value to caller so no copy is made for returned value. If you make return variable `const` the compiler cannot perform move and is forced to return the value by copy. It is called return value optimization.

I didn't ask you what is "return value optimization". I said: "const improves readability and Qt has Implicit Sharing".

I'd say I have wasted more time handling this issue than the "performance improvement" applied to even millions of users.

ouuan:
> This PR should be merged today, because if we develop on other branches, this will inevitably cause a cascade of merge conflicts.

I said that I would have very limited time on developing 3 months ago. Now when I'm not in front of a computer I sometimes list the things I want to do. The list is getting longer and longer.

The development of CP Editor can slow down, but I can't merge your PRs without reviewing them. So I'll be pleased if you can slow down. And I hope most times I can only give suggestions but not try to find solutions because I don't have much time.

coder3101:
Here is what I have suggest.

1. We will make use of CMakeList.txt for linting. (It has some other benfits too, like if you use clangd, the LSP can lint clang-tidy reports in your IDE itself when you are coding. It will be possible to run the check in different OS too)
2. I will create a new action say "clang-tidy-checks", It will be a clone of `build_test.yml` except that it will run with `cmake ..  -DUSE_CLANG_TIDY=On`, It will only run on PR.

This way if normal build test passes in 2min or so, you can be sure that there was no compilation error. If this action fails in 4-5 min, it means clang-tidy reported some errors. Currently, Codeacy also takes over 5 min to report issues.

coder3101:
> > This PR should be merged today, because if we develop on other branches, this will inevitably cause a cascade of merge conflicts.
> 
> I said that I would have very limited time on developing 3 months ago. Now when I'm not in front of a computer I sometimes list the things I want to do. The list is getting longer and longer.
> 
> The development of CP Editor can slow down, but I can't merge your PRs without reviewing them. So I'll be pleased if you can slow down. And I hope most times I can only give suggestions but not try to find solutions because I don't have much time.

You can take your time. I can understand, in one month, I will also have an internship and will not get much time except in weekends.

src/Settings/DefaultPathManager.cpp
coder3101: I'll disable this check.

ouuan:
>  2. It will be a clone of `build_test.yml` except that it will run with `cmake ..  -DUSE_CLANG_TIDY=On`, It will only run on PR.

Clone is not a good idea, which makes modification hard and easy to make mistakes. You can try [run-clang-tidy.py](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py), which is at `/usr/share/clang/run-clang-tidy.py` on my Arch Linux (in the `clang` package`). However, I'm not sure which solution is the best.

coder3101:
> You can try [run-clang-tidy.py](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py), which is at `/usr/share/clang/run-clang-tidy.py` on my Arch Linux (in the `clang` package`).

This just helps in running clang-tidy in parallel. It does not solves issue of having all headers and up-to-date compile-commands.json. If you wish to run clang-tidy separately, you have to pass the exact same compiler flags to it as you pass to a compiler. For this purpose we export compile-commands.json, which can only be exported if `cmake ..` runs which requires Qt to be installed. So my point is even if you create a whole new workflow for clang-tidy, you still need to install Qt, it has to run on all 3 OS, to cover all code. This is already 30% of `build_test.yml`. 

However I agree with your opinion of not cloning because of redundancy, So why dont we not use cache for this new action. It will result in very small workflow file with just these steps:
```
0. Run on Windows/Linux/MacOS
1. Install Qt and clang-tidy
2. Run `cmake .. -DUSE_CLANG_TIDY=On`
3. Run cmake --build
```
I think this should complete in under 5 min and it is small so easy to maintain.

coder3101:
Clang tidy checks are being completed in about 5 mins. This should be acceptable.

ouuan (approved):
LGTM

coder3101:
Please remove Codacy Static analysis integration from this repository.

PR #708, opened 2020-12-12
docs: add GitHub Discussions

ouuan:
The Telegram group will be mentioned only on the website, and the Slack workspace will be deprecated.

The badges are removed because now there are only "downloads@all" and "help wanted issues", which are not very helpful.

@IZOBRETATEL777 you can update the README.


coder3101 (approved):
LGTM

PR #784, opened 2021-03-06
ci: separate check translations, merge build and clang-tidy test

ouuan:
1. The separated workflow for translations makes it easier to realize that it's the translations that are not
up to date, instead of the build failed.
2. The build test was somehow a duplicate of the Clang-Tidy check.


coder3101 (approved):
LGTM. Much cleaner

PR #808, opened 2021-04-02
chore(donators): add GoMage

coder3101:
Add GoMage https://www.gomage.com/ for their contribution of $1 toward cpeditor.

### DIFF
--- a/DONATORS.md
+++ b/DONATORS.md
@@ -4,7 +4,7 @@ None.
 
 ## Open Collective
 
-None.
+- [GoMage](https://www.gomage.com/)
 
 ## WeChat
 


ouuan:
What about including the amount of money?

coder3101:
> What about including the amount of money?

Why take the task of keeping the list updated like in case of recurring donation and people can get their names in this list even for $1, that's the reward they get, as for actual amount we should let open collective or others keep track of.

ouuan (approved):
I agree that the need for updating is a drawback. The advantage is to show users that "people can get their names in this list even for $1".

However, I'm OK with either including the amount of money or not.

PR #849, opened 2021-06-02
feat: support opening source files by clicking them on macOS

### LINKED ISSUE #846: Associate source files to open in cpeditor on macOS

CPEditor doesn't register itself as a cpp file open method , so I can't open cpp files via CPEditor in my file manager.

Register CPEditor as a cpp file open method.

**Additional context**

swift-zym:
## Description

Listen `QFileOpenEvent` to open files and register as an editor for code files on macOS.

Closes #846 

Untitled

## Checklist


src/application.hpp
```diff
@@ -0,0 +1,34 @@
...
+#ifndef APPLICATION_HPP
+#define APPLICATION_HPP
+
+#include <QFileOpenEvent>
+#include <singleapplication.h>
+
+class Application : public SingleApplication
+{
+  public:
+    Application(int &argc, char **argv) : SingleApplication(argc, argv, true)
```
sonulohani, line 27: Use explicit before ctor. Move the implementation to cpp file.

src/appwindow.hpp
```diff
@@ -57,6 +57,8 @@ class AppWindow : public QMainWindow
 {
     Q_OBJECT
 
+    friend class Application;
```
sonulohani, line 60: Dont use friend class. This will break the oops principles. Use setter getter if you want any attribute access from the class

src/application.hpp
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+
+#ifndef APPLICATION_HPP
+#define APPLICATION_HPP
+
+#include <QFileOpenEvent>
+#include <singleapplication.h>
+
+class Application : public SingleApplication
```
sonulohani, line 24: Whats the intention of bringing new class. Can't you use existing SingleApplication class?

src/application.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "Core/EventLogger.hpp"
+#include "appwindow.hpp"
+
+bool Application::event(QEvent *event)
+{
+    if (event->type() == QEvent::FileOpen)
+    {
+        auto *openEvent = static_cast<QFileOpenEvent *>(event);
+        auto file = openEvent->file();
+        LOG_INFO("Open file : " << file);
```
sonulohani, line 32: Should be "Opening file"

src/application.cpp
```diff
@@ -0,0 +1,41 @@
...
+        auto widgets = QApplication::topLevelWidgets();
+        for (auto *widget : widgets)
+        {
+            auto *appWindow = qobject_cast<AppWindow *>(widget);
```
sonulohani, line 36: QApplication is singleton class. Just use `QApplication::instance()` to get the QApplication object. No need to iterate over all the objects.

sonulohani (changes requested):

src/application.hpp
coder3101: explicit is required for constructors taking only 1 non default parameter

See:  https://en.cppreference.com/w/cpp/language/explicit

src/application.hpp
coder3101: How to Override event method without inheriting?

sonulohani: > How to Override event method without inheriting?

Are you asking me?

coder3101: I mean there was no other way around except inherit to override

coder3101: See this: https://doc.qt.io/qt-5/qfileopenevent.html

sonulohani: I didnt know that there is no singleapplication.hpp file. I thought we could reuse SingleApplication class. 😀😀😀

coder3101:
From QFileOpenEvent docs

> Note: This class is currently supported for macOS only.

I expected some OS macros, how is it being handled in other OS?

src/application.hpp
sonulohani: that is until c++11. Its always good practice to have explicit before ctor

coder3101: If static analysis passed it's okay.

src/application.cpp
coder3101:
```suggestion
        LOG_INFO("Opening file : " << file);
```

src/application.cpp
coder3101: `openTab` is a member of applicationwindow.

sonulohani (dismissed):
LGTM. Wait for others also to review the code before merging it into master.

ouuan (approved):
Now it looks good to me. However, we still need someone to test it on macOS.

src/application.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "appwindow.hpp"
+#include <QFileOpenEvent>
+
+Application::Application(int &argc, char **argv) : SingleApplication(argc, argv, true)
+{
+}
+
+bool Application::event(QEvent *event)
+{
+#ifdef Q_OS_MAC
```
ouuan, line 29: Do we really need `#ifdef Q_OS_MACOS`? This class currently supports macOS only, but it's harmless on unsupported platforms.

ouuan: BTW, `Q_OS_MAC` is deprecated, and we should use `Q_OS_MACOS`.

swift-zym: I can't expect what will this class do on other OSs,so I added these macros.

ouuan: > I can't expect what will this class do on other OSs

It should do the same thing. When it's unsupported, I think this event won't be triggered.

coder3101:
> Now it looks good to me. However, we still need someone to test it on macOS.

I will test and let you know.

ouuan (approved):

coder3101 (approved):
Looks good and works fine as well.

swift-zym:
BTW, should we also do this for Windows (by editing the registry files)?

ouuan:
> BTW, should we also do this for Windows (by editing the registry files)?

Oh, I forgot that this is just a part of #846. We can do it for both Windows and Linux.

ouuan:
I think we'd better open a new issue. #846 was two problems mixed into one issue.

swift-zym:
> I think we'd better open a new issue. #846 was two problems mixed into one issue.

I've open #855.

PR #897, opened 2021-06-28
feat: watch checker source file

### LINKED ISSUE #843: Support update/remove custom checkers

Now cpeditor doesn't auto re-compile custom checker when checker's source file changed.

Provide "Update" and "Delete" buttons next to "Add Checker".

**Additional context**

swift-zym:
## Description

Ask whether to re-compile current checker when it changed.

Fixes #843 
## How Has This Been Tested?

macOS 12.0
## Checklist


ouuan:
Do we need a prompt and an option? I think a silent recompilation could be the only behavior :thinking:

swift-zym:
> Do we need a prompt and an option? I think we can silently recompile and don't need an option 🤔

I think this will cause the checker can't be use if the source file is deleted

ouuan:
> I think this will cause the checker can't be use if the source file is deleted

Then just don't recompile when it's deleted.

ouuan:
I think it's not very good to recompile on file changes, whether prompted or not, if you are editing and saving the checker.

ouuan:
Maybe we should check this on execution.

PR #910, opened 2021-07-08
Add version info for windows

neko-para:
Add version.rc


coder3101 (approved):

ouuan:
Can we use DISPLAY_VERSION here? :thinking: (not familiar with version.rc)

neko-para:
> Can we use DISPLAY_VERSION here? 🤔 (not familiar with version.rc)

I'm not sure. There are two types. One is where we use now, which should be 4 numbers. The othere is in `StringFileInfo`, which allows a string. But the properties dialog show the first version. I don't know where does the string version used.

neko-para:
Ref: [VERSIONINFO](https://docs.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource)

ouuan (approved):

coder3101:
> Can we use DISPLAY_VERSION here? :thinking: (not familiar with version.rc)

I think the version should be sem ver across all packages and release types like alpha/beta/stable.

CMAKE allows sem version of type: `Major.Minor.Patch.Tweak`. Tweak can be omitted if its zero. Then we can have only 1 version throughout the project. i.e PROJECT_VERSION.

Ex:
For stable : 6.8.2 (Just like old)
For beta : 6.9.2 

For alpha/master : 6.9.2.23 
(tweak is revision since 6.9.2, last beta)

In cmake project version, we still keep 6.8.2 but check if it's not tagged we set tweak to revision count.

In this way, For master we don't need DISPLAY_VERSION to distinguish it and by sem version: stable < beta < master is always followed.

Snap edge channel uses this type of sem versioning, I think AUR and others should use the same.

ouuan:
The commit sha is useful, so we should use DISPLAY_VERSION when applicable. We can use SNAP_VERSION for version.rc on Windows.

coder3101:
> The commit sha is useful, so we should use DISPLAY_VERSION when applicable. We can use SNAP_VERSION for version.rc on Windows.

Fine, then `SNAP_VERSION` should be renamed to something like `PROJECT_VERSION_FULL`

PR #932, opened 2021-07-15
ci(snap): fix version and update README

coder3101:


ouuan:
The "Get it from" is invisible with a height of 28px. Do we need that line?

coder3101:
> The "Get it from" is invisible with a height of 28px. Do we need that line?

Not necessary but why do we keep it along with downloads, sponsors and other badges. I suggest we put it in a new line below others, it is visible and centered to draw focus of users.

coder3101:
Something like this:
Screenshot from 2021-07-15 09-51-07

ouuan:
Yeah, but I'm wondering why we should draw attention to it. The badge somehow sounds like "you need to get it from snap store", but actually the snap installation method is targeted at only a few users.

coder3101:
We set the `width=15%` and use the second row in README for listing badges with store presence. 
Screenshot from 2021-07-15 10-42-45

ouuan:
I think the badges in the version table are enough.

coder3101:
> I think the badges in the version table are enough.

Fine, remove the "Get it from Snap store" badge and approve.

ouuan (approved):

neko-para (approved):

ouuan:
I reverted the "fix version" part by mistake...

ouuan:
> I reverted the "fix version" part by mistake...

However, it's a good example that we should do one thing per commit.

PR #858, opened 2021-06-05
refactor: rewrite settings to support object

neko-para:
update genSettings.py, SettingsInfo.xxx to support Object and QMap:xxx.

## Description

I restructured the valuewrapper, so that it can support recursive widget.
Also, I add a SettingIter to correctly locate the SettingInfo, with extra infomation(the key of object or qmap),

## Related Issues / Pull Requests

#532 

## Checklist

Check the old codes to see how to implement some of these:

- [x] The items on the pages should be placed in the center.
- [x] QPlainTextEdits on the pages should fill the whole page instead of having a height limit.
- [x] The "(?)" should link to translated anchors on the website. (e.g. it should be https://cpeditor.org/zh/docs/preferences/code-edit/#%E8%87%AA%E5%8A%A8%E7%BC%A9%E8%BF%9B instead of https://cpeditor.org/zh/docs/preferences/code-edit/#auto-indent)
- [x] Settings shouldn't need a manual transfer, if the old version is v6.8.
- [x] Snippets shouldn't need a manual transfer.
- [x] Default Language should be moved to the same page of Language Config.
- [x] Documentation should be refactored in the same structure. (Links to the website should work.)
- [x] Releases should contain/require yaml-cpp.

PR #939, opened 2021-07-27
add script to build a fake DEB

neko-para:


ouuan:
No explanation for this? Why add it?

neko-para:
In #858 we adds a new requirement (libyaml-cpp), and I've noticed that there's not script to build the deb. (Other releases as well)
So I copy the commands from release.yaml to this script with some adjustment, and upload it in case it would be used in the future.

ouuan:
If we have to add a script, why not reuse it in the workflow? But do we really need this script :thinking:? We can add a workflow to test DEB build, or trigger a test build in a fork.

neko-para:
Then I will remove that.

PR #1009, opened 2021-11-10
feat: add stopwatch

### LINKED ISSUE #1005: Add stopwatch
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this feature provided in the master branch?

- [X] I have checked the changelog

### The problem you are facing

I usually keep track of the time taken to solve the problem. I think a stopwatch would be a useful feature.

### Describe the feature you'd like

- Add an option in settings to make the stopwatch visible in the right panel.
- If the option is enabled then display the stopwatch in the right panel with the start/stop and reset buttons.

<details>
<summary>It looks something like this.</summary>

Screenshot_20211104_223101
</details>

### Describe alternatives you've considered

N/A

### Are you willing to contribute to this?

- [X] I'm ready to contribute to this (and will open a PR soon)
- [ ] I'd like to have a try (with the help of the maintainers)
- [ ] No, thanks

### Anything else?

Discussion: https://github.com/cpeditor/cpeditor/discussions/1004

hitanshu-mehta:
## Description

- Added `Stopwatch` class using `QElapsedTimer` which emits a `time` signal every second. It also emits a `StateChanged` signal whenever the state of the stopwatch is changed.
- These signals are connected with appropriate slots using which `StopwatchWidget` is updated.
- Two new options are added in `Preferences/Actions/Stopwatch`.
  - Display Stopwatch
  - Automatically start/stop stopwatch on tab switch

fixes #1005 

## Motivation and Context

The stopwatch will be useful to someone who wishes to keep track of the time taken to solve the problem.

## How Has This Been Tested?

ManjaroLinux - 21.1.6

<details>
<summary>It looks something like this.</summary>

Screenshot_20211104_223101
</details>

## Checklist

## Additional text

- Translations need to be updated.


hitanshu-mehta:
MacOs build failure seems unrelated 🤔

coder3101:
> MacOs build failure seems unrelated 🤔 

Clang tidy reported some suggestions that you should either fix or if its false alarm, Suppress it with an inline comment.

ouuan:
macOS uses the latest version of clang-tidy, so it's broken now.

See https://github.com/cpeditor/cpeditor/pull/999#issuecomment-944371434. CP Editor is in slow development now, so I planned to wait for Arch Linux's clang 13 and resolve it at local.

ouuan:
However, anyone is welcome to fix the clang-tidy issue, while I'm waiting for clang 13 on Arch. It should be fixed in a separate pull request.

hitanshu-mehta:
PR is ready for review :)

src/Core/Stopwatch.hpp
```diff
@@ -0,0 +1,119 @@
...
+     * @brief reset to stopwatch to inactive state.
+     */
+    void reset();
+
+  signals:
+    /**
+     * Emits a signal with the current stopwatch time.
+     * @param t Current stopwatch time
+     */
+    void time(int t);
```
ouuan, line 93:
```suggestion
    void timeChanged(int ms);
```

src/mainwindow.hpp
```diff
@@ -157,6 +160,9 @@ class MainWindow : public QMainWindow
     void updateChecker();
     void runTestCase(int index);
 
+    void updateStopwatch(int time);
```
ouuan, line 163:
```suggestion
    void updateStopwatch(int ms);
```

src/mainwindow.cpp
```diff
@@ -1445,3 +1507,18 @@ void MainWindow::onRunKilled(int index)
...
 }
+
+// --------------------- Stopwatch SLOTS ----------------------------
+
+void MainWindow::updateStopwatchButtons(Core::Stopwatch::State state)
+{
+    ui->stopwatchResetButton->setDisabled(stopwatch->isInactive());
+
+    stopwatch->isRunning() ? ui->stopwatchStartStopButton->setText("Stop")
+                           : ui->stopwatchStartStopButton->setText("Start");
```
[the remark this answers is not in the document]
ouuan: And we can use the `state` parameter instead of members of `stopwatch`.

ouuan (changes requested):

ouuan:
And I prefer making `Stopwatch` a widget instead of just an object.

src/mainwindow.hpp
```diff
@@ -248,5 +258,9 @@ class MainWindow : public QMainWindow
     QString compileCommand() const;
     int timeLimit() const;
     void updateCompileAndRunButtons() const;
+    void setStopwatch();
+
+    virtual void hideEvent(QHideEvent *event) override;
+    virtual void showEvent(QShowEvent *event) override;
```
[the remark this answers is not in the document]
hitanshu-mehta: `onTabChanged` signal will only give the index of the currently changed tab. But we need to stop the stopwatch of the previous tab, before switching to a new tab. I was not able to find a way to achieve this only using `onTabChanged` signal.

src/Core/Stopwatch.hpp
```diff
@@ -0,0 +1,119 @@
...
+        Running
+    };
+
+    explicit Stopwatch(QObject *parent = nullptr);
+
+    /**
+     * @brief set the stopwatch refresh granularity
+     * @param g The granularity to be set
+     */
+    void setGranularity(Granularity g);
```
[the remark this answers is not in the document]
hitanshu-mehta: Yes. sorry I will remove it.

src/mainwindow.hpp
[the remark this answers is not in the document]
coder3101: I believe the objective is: current visible tab should have active timer and others should have paused timer.

You could achieve it in 3 ways:

1. Add a member function to keep track of last tab, I believe in this case when restoring many tabs  on startup it will become more complex. 

2. Have a single Stopwatch attached to app window with multiple states for each tabs. 

3. On every tab switch iterate all other tabs and stop the timer if running except the current tab. Simplest but could be performance intensive (needs benchmark against current implementation).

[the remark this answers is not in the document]
hitanshu-mehta: I think that the current implementation is simple that all 3 ways you suggested. Is there any reason for not preferring this solution?

src/mainwindow.cpp
```diff
@@ -1445,3 +1507,18 @@ void MainWindow::onRunKilled(int index)
...
+}
+
+void MainWindow::updateStopwatch(int time)
+{
+    ui->stopwatch->setText(QTime(0, 0).addMSecs(time).toString("hh:mm:ss"));
+}
```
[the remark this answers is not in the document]
hitanshu-mehta: `QTime(0, 0, 0, ms)` here ms should be in the range of 0 to 999 (https://doc.qt.io/qt-5/qtime.html#QTime-2), which won't be always the case.

hitanshu-mehta:
> And I prefer making Stopwatch a widget instead of just an object.

@ouuan `Stopwatch` is now the widget. `MainWindow` doesn't have the responsibility of updating the UI of the stopwatch now.

src/Widgets/Stopwatch.hpp
```diff
@@ -0,0 +1,121 @@
...
+    bool isInactive() const;
+
+  private:
+    enum class Granularity : int
+    {
+        Milliseconds = 1, // Stopwatch refreshed every msec
+        Hundredths = 10,  // Stopwatch refreshed every 10 msec
+        Tenths = 100,     // Stopwatch refreshed every 100 msec
+        Seconds = 1000    // Stopwatch refreshed every sec
+    };
```
ouuan, line 89: Do we really need this :thinking:?

src/Widgets/Stopwatch.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * The UI structure:
+     *
+     * mainLayout:
+     *  - timeLabel
+     *  - startPauseButton
+     *  - resetButton
+     */
+    QLayout *mainLayout = nullptr;
```
ouuan, line 106: `mainLayout` can be declared as a local variable in the constructor.

src/mainwindow.cpp
```diff
@@ -1300,6 +1315,34 @@ void MainWindow::updateCompileAndRunButtons() const
...
+
+void MainWindow::showEvent(QShowEvent *event)
+{
+    if (event->spontaneous())
+    {
+        QWidget::showEvent(event);
+        return;
+    }
+
+    if (SettingsHelper::isStartStopStopwatchOnTabSwitch() && stopwatch->isPaused())
```
ouuan, line 1340:
```suggestion
    if (SettingsHelper::isStartStopStopwatchOnTabSwitch() && !stopwatch->isRunning())
```

src/Widgets/Stopwatch.cpp
```diff
@@ -0,0 +1,155 @@
...
+    update();
+}
+
+void Stopwatch::start()
+{
+    if (isInactive())
+    {
+        accumulator = 0;
+        elapsedTimer.start();
+        startTimer(getGranularity());
```
ouuan, line 73: What does this do :thinking:

src/Widgets/Stopwatch.cpp
```diff
@@ -0,0 +1,155 @@
...
+    resetButton = new QPushButton("Reset");
+
+    // set up UI
+    mainLayout->addWidget(timeLabel);
+    mainLayout->addWidget(startPauseButton);
+    mainLayout->addWidget(resetButton);
+
+    // set up timer
+    timer = new QTimer(this);
+    timer->setInterval(getGranularity());
```
ouuan, line 48: I suggest using a single-shot timer and setting `1000 - accumulator % 1000` as the interval on each update.

ouuan (changes requested):

src/Widgets/Stopwatch.cpp
hitanshu-mehta: Sorry. I forgot to remove it. I was trying out something. Not needed now.

src/Widgets/Stopwatch.cpp
hitanshu-mehta: Sorry. I didn't get the reason for doing this. Can you please explain?

src/Widgets/Stopwatch.hpp
hitanshu-mehta: Not really, I will remove it.

src/Widgets/Stopwatch.cpp
ouuan: > Sorry. I didn't get the reason for doing this. Can you please explain?

Now the time updates every 1 second, so it can be updated at either 1.000 or 1.999.

ouuan: And maybe we can use `1001 - accumulator % 1000`/`1010 - accumulator % 1000`, etc. to avoid multiple timeouts per second.

ouuan: I discovered this because I observed that it shows 00:00:01 when the timer has started for almost 2 seconds.

ouuan: > I discovered this because I observed that it shows 00:00:01 when the timer has started for almost 2 seconds.

Haven't tried the solution above, so I'm not sure if I got the correct cause for it.

hitanshu-mehta: `1001 - accumulator % 1000/1010 - accumulator % 1000` if we use this formula then the update interval will always be 0. 🤔 

> I discovered this because I observed that it shows 00:00:01 when the timer has started for almost 2 seconds.

If this is the problem then reducing the interval to something smaller like 500ms will help.

ouuan: > If this is the problem then reducing the interval to something smaller like 500ms will help.

I think my solution sounds better :thinking: At least, 500ms is not enough. Reducing to 10ms or even smaller could be much better if it doesn't affect performance and the accuracy of time measurement.

hitanshu-mehta: Let's try out your solution 😃 !

ouuan:
Any progress?

hitanshu-mehta:
Sorry, I'm taking time 😅 . I was busy with some other stuff. I will start working on the feature this weekend.

hitanshu-mehta:
@ouuan I have started using the SingleShot timer `granularity + 10 - accumulator
  % granularity`. Here are a few things I observed.
1. UI updates are not happening smoothly due to different interval values each time we pause. 
2. I also experimented with different values of the interval with the previous implementation (i.e. with a regular timer), values between `300ms` to `400ms` seem good enough to me.

Please try out this change and let me know your thoughts :)

src/Settings/settings.json
```diff
@@ -1208,5 +1208,22 @@
...
+    "name": "Display Stopwatch",
+    "type": "bool",
+    "default": false
+  },
+  {
+    "name": "Toggle Stopwatch On Tab Switch",
+    "desc": "Start/Stop stopwatch on tab switch",
+    "type": "bool",
+    "default": false,
+    "tip": "When switching to a different tab, automatically start the stopwatch\non the current tab and pause the stopwatch on the previous tab.",
```
hitanshu-mehta, line 1222:
```suggestion
    "tip": "When switching to a different tab, automatically start the stopwatch on the current tab and pause the stopwatch on the previous tab.",
```

ouuan: The tooltip doesn't auto-wrap. Break the line to prevent it from being too long.

src/Widgets/Stopwatch.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    update();
+}
+
+void Stopwatch::reset()
+{
+    elapsedTimer.invalidate();
+    accumulator = 0;
+    timer->stop();
+    QCoreApplication::processEvents();
```
ouuan, line 105: Can you explain this? I think it would be better to add a comment here.

src/Settings/settings.json
hitanshu-mehta: Sorry my bad 🤦🏻‍♂️

src/Widgets/Stopwatch.cpp
hitanshu-mehta: I added that to prevent GUI from freezing. But I think we don't need it because we are not doing any long computations that can freeze GUI. Hence, I will remove it.

ouuan (approved):
LGTM, thanks for your contribution!

ouuan:
@allcontributors add @hitanshu-mehta as a contributor for ideas and code.

PR #1031, opened 2022-01-01
feat: support stress testing

### LINKED ISSUE #917: Stress testing
This was moved to Discussions at #162 by mistake.

<div type='discussions-op-text'>

<sup>Originally posted by **ouuan** February  6, 2020</sup>

I can't use stress testing inside the editor.

Add a window for stress testing, it has two (or three, including the current solution, maybe sync with the main editor) editors — one for the other solution and one for the test case generator, you can load them from files, maybe load from other tabs. You can set the time limit, and run until finding a diff between the two solutions. The number of correct test cases and used time for each test case can be displayed while stress testing.

N/A

**Additional context**

N/A
</div>

swift-zym:
## Description

Support stress testing.

Closes #917 

## How Has This Been Tested?

macOS 12.0.1

https://user-images.githubusercontent.com/61449077/147852704-43e01e3f-bf33-4558-9f6c-61eebd5f75af.mov

## Checklist

- [ ] If needed, I have opened a pull request or an issue to update the [documentation](https://github.com/cpeditor/cpeditor.github.io).

## Additional text
It is almost the same as [Polygon](https://polygon.codeforces.com)'s, so only testilb style generators are supported.

### DIFF
--- a/src/Widgets/StressTesting.cpp
+++ b/src/Widgets/StressTesting.cpp
@@ -138,9 +138,9 @@ void StressTesting::start()
             if (!ok)
                 break;
 
-            argumentsRange.append(qMakePair(argumentsCount == 1 ? left - 1 : left, right));
+            argumentsRange.append(qMakePair(left, right));
 
-            currentValue.push_back(left);
+            currentValue.push_back(argumentsCount == 1 ? left - 1 : left);
 
             leftBracketPos = -1;
         }
--- a/src/Widgets/StressTesting.cpp
+++ b/src/Widgets/StressTesting.cpp
@@ -1,0 +1,487 @@
+/*
+ * Copyright (C) 2019-2021 Ashar Khan <ashar786khan@gmail.com>
+ *
+ * This file is part of CP Editor.
+ *
+ * CP Editor is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * I will not be responsible if CP Editor behaves in unexpected way and
+ * causes your ratings to go down and or lose any important contest.
+ *
+ * Believe Software is "Software" and it isn't immune to bugs.
+ *
+ */
+
+#include "Widgets/StressTesting.hpp"
+#include "Core/Checker.hpp"
+#include "Core/Compiler.hpp"
+#include "Core/EventLogger.hpp"
+#include "Core/MessageLogger.hpp"
+#include "Core/Runner.hpp"
+#include "Settings/PathItem.hpp"
+#include "Util/FileUtil.hpp"
+#include "Widgets/TestCase.hpp"
+#include "generated/SettingsHelper.hpp"
+#include "mainwindow.hpp"
+#include <QCodeEditor>
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QLineEdit>
+#include <QPushButton>
+#include <QTemporaryDir>
+#include <QVBoxLayout>
+#include <functional>
+
+namespace Widgets
+{
+StressTesting::StressTesting(QWidget *parent)
+    : QMainWindow(parent), mainWindow(qobject_cast<MainWindow *>(parent)), compiledCount(0), runFinishedCount(0),
+      argumentsCount(0), stopping(false)
+{
+    log = mainWindow->getLogger();
+
+    auto *widget = new QWidget(this);
+    auto *layout = new QVBoxLayout();
+    widget->setLayout(layout);
+    setCentralWidget(widget);
+    setWindowTitle(tr("Stress Testing"));
+    resize(400, 360);
+
+    auto *generatorLayout = new QHBoxLayout();
+    generatorLable = new QLabel(tr("Generator Path:"), widget);
+    generatorLayout->addWidget(generatorLable);
+    generatorPath = new PathItem(PathItem::CppSource, widget);
+    generatorLayout->addWidget(generatorPath);
+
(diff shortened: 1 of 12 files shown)


src/Widgets/StressTesting.cpp
```diff
@@ -0,0 +1,487 @@
...
+    if (exitCode == 0)
+    {
+        log->info(head, tr("Execution has finished in %1ms").arg(timeUsed));
+        if (index == 0)
+        { // Generator
+
+            in = out;
+
+            QString language = mainWindow->getLanguage();
+            stdRunner = new Core::Runner(1);
```
coder3101, line 388: You can modify or inherit from Runner a class `StressRunner` and move these logic inside it. That way you can also change enum types.

src/Widgets/StressTesting.cpp
```diff
@@ -0,0 +1,487 @@
...
+}
+
+void StressTesting::onStdCompilationFinished()
+{
+    log->info(tr("Compiler"), tr("Standard program compilation has finished"));
+    compiledCount++;
+    if (compiledCount == 3)
+    {
+        nextTest();
+    }
```
coder3101, line 282: Could there be a better way for it?

coder3101 (commented):
Awesome! Left some comments please provide your feedback

src/Widgets/StressTesting.cpp
swift-zym: Why do we need a new class?

coder3101: If you make SourceType as enum class it cannot be implicitly converted to int type. Runner takes an int argument hence to make it work you might have to inherit it. Since you used normal enum its fine.

coder3101:
I will let ouuan take a look at it since he reviewed it before as well.

ouuan:
Just take a quick look at the initial dialogs:

- Does it support selecting unsaved files as generator/std?
- Add counterexample to testcase should be a button after it's found, not an option before starting.

swiftqwq:
@ouuan 

- I've added a progress bar and a preview for counterexamples.
- Separators are now added after `%{index}` to avoid that problem, and I've fixed some other issues with pattern parsing.
- I've reduced the number of messages for each generated case, but kept two messages so that users can check whether the pattern works as they expect.

ouuan:
- Now the user cannot stop the stress testing without saving a counterexample to the test cases if counterexamples are found very quickly. I suggest making "stop testing" a checkbox: [add button] [ignore button] (stop checkbox).
- I think the "Accepted" message can be removed. A message can be displayed for saving a test case.
- I think the current place of the action "new file from generator template" is not very convenient/intuitive. I suggest removing it from the menu (or moving it to beside stress testing 🤔), and adding a button next to the generator/std path selection to create a new file, where the generator template or a copy of the current code is used, respectively.

ouuan:
Well, if we keep the "new file from template" submenu, we can also add options for C++/Java/Python.

Another problem is that the progress should be 100% instead of 0% when all tests are finished, and it should stop on file not found.

ouuan:
I'm thinking that the generator argument design might be unintuitive for users unfamiliar with the testlib-style generator / generator with arguments. Maybe we can have two modes, one without arguments and one with an arguments pattern. In the no argument mode, I'm not sure whether it's better to set the number of test cases, to set the total running duration, or to run indefinitely until manual stop :thinking:

This can be simulated by `[1..n]`, but the user might not know this without hints. And maybe it's just simpler to run until manual stop. I think that's what I always did with a simple bash script.

translations/zh_CN.ts
```diff
@@ -2710,12 +2758,40 @@ This may reduce distractions caused by stopwatch updates.</source>
     </message>
     <message>
         <source>Enable CF Tool</source>
-        <translation type="unfinished"></translation>
+        <translation type="unfinished">CF Tool 运行出错</translation>
```
ouuan: incorrect and unfinished

translations/zh_CN.ts
```diff
@@ -2968,6 +3044,165 @@ This may reduce distractions caused by stopwatch updates.</source>
...
+        <source>Stop stress testing</source>
+        <translation>停止对拍</translation>
+    </message>
+    <message>
+        <source>Counterexample added to testcases</source>
+        <translation>会产生错误的测试用例已被添加</translation>
+    </message>
+    <message>
+        <source>Use Generator Arguments Pattern:</source>
+        <translation>使用数据生成器参数模式：</translation>
```
ouuan: “模式”这个翻译本来就有点歧义，在“使用”这个语境下更奇怪了🤔我个人倾向于不翻译写“pattern”，或者直接删掉，“使用数据生成器参数”

就算是英文这个，感觉也有点奇怪，因为这个是选择是否使用 argument，而非是否使用 pattern，或许原文就可以删掉“pattern”

translations/zh_CN.ts
```diff
@@ -2968,6 +3044,165 @@ This may reduce distractions caused by stopwatch updates.</source>
...
+        <source>Select generator from editor tabs...</source>
+        <translation>从编辑器界面中选择数据生成器...</translation>
+    </message>
+    <message>
+        <source>Select standard program from editor tabs...</source>
+        <translation>从编辑器界面中选择标准程序...</translation>
+    </message>
+    <message>
+        <source>Counterexample Found</source>
+        <translation>找到会产生错误的测试用例</translation>
```
ouuan:
```suggestion
        <translation>发现的反例</translation>
```

swiftqwq: 这里改成“发现反例”是不是好一点

ouuan: 确实，之前没看上下文理解错了

PR #1087, opened 2022-06-16
ci(snap): specify commit hash for action-publish

ouuan:


coder3101:
Why not `master`? If snap team makes some change which might be handled by action update it will fail for us since we are always using fixed action version.

A drawback of using master could be that it can potentially break any anytime if action itself brings some breaking change in its yaml schema.

ouuan:
Maybe we can simply wait for a release. https://github.com/snapcore/action-publish/issues/28 is opened yesterday so it will probably get fixed soon.

ouuan:
It has been released: https://github.com/snapcore/action-publish/commit/0a8d537ae06f4a292e8b4ef1084cd5631b3c6871

But now we get a different error message: https://github.com/cpeditor/cpeditor/runs/7023348369?check_suite_focus=true

ouuan:
It works now.

PR #1155, opened 2023-03-20
build: fix build on Arch Linux with CMake 3.26

ouuan:
## Description

It seems that a recent update of CMake breaks the build on Arch: https://github.com/cpeditor/cpeditor/actions/runs/4456830088

I'm opening this PR to test it against ALA on 3.13 first, and then to see how to fix it.

## How Has This Been Tested?

I downgraded CMake on my laptop and the build succeeded.

### DIFF
--- a/.github/workflows/aur-test.yml
+++ b/.github/workflows/aur-test.yml
@@ -46,6 +46,7 @@ jobs:
         id: build-pkg
         with:
           path: ${{ matrix.pkg.path }}
+          archive: 2023/03/13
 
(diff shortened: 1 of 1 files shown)


ouuan:
Fixed in https://github.com/cpeditor/QCodeEditor/pull/42 and https://github.com/cpeditor/lsp-cpp/pull/4. I'm closing this because it should be backported to v6.10, and it will be fixed in master by merging v6.10.

PR #1166, opened 2023-04-09
fix: make signal handler more robust & fix clang-tidy issues

ouuan:
- Disable `cppcoreguidelines-avoid-do-while`
- Use C++ 17 to fix `bugprone-signal-handler`


swift-zym (dismissed):

ouuan:
I'm not sure whether using C++17 is a real fix for `bugprone-signal-handler`. Maybe we really need to fix the signal handler 🤔

ouuan:
Maybe we should follow [Calling Qt Functions From Unix Signal Handlers | Qt 5.15](https://doc.qt.io/qt-5/unix-signals.html)

ouuan:
Tested on Arch Linux only.

coder3101 (approved):

PR #934, opened 2021-07-18
breaking: change app name and log path

ChungZH:
## Description

Change the log from `/tmp/cpeditorLogFiles` to `~/.cache/cpeditor/log`.

Use [QStandardPaths::](https://doc.qt.io/qt-6/qstandardpaths.html)CacheLocation.

Also change the appname from CP Editor to cpeditor.

## Related Issues / Pull Requests

close #926

## How Has This Been Tested?

Archlinux.
## Checklist

- [x] If the key of a setting is changed, the `old` attribute is updated or it is resolved in SettingsUpdater.
- [x] If there are changes of the text displayed in the UI, they are wrapped in `tr()` or `QCoreApplication::translate()`.
- [ ] If needed, I have opened a pull request or an issue to update the [documentation](https://github.com/cpeditor/cpeditor.github.io).
- [ ] If these changes are notable, they are documented in [CHANGELOG.md](https://github.com/cpeditor/cpeditor/blob/master/CHANGELOG.md).

Issue #769, opened 2021-02-13
Breaking changes for CP Editor 7.0 in the future


ouuan:
There could be some changes that break something, especially the settings.

We can list them here, and apply them when we are ready for CP Editor 7.0.

~~- [ ] Drop support for the current editor themes. (#516)~~ Now KSyntaxHighting supports [many themes](https://github.com/KDE/syntax-highlighting/tree/master/data/themes).
- [ ] #925
- [ ] #926
- [ ] #927
- [ ] #928

coder3101:
Closed, Individual issues are created and marked with v7.0 milestone.

PR #1308, opened 2024-02-13
chore!: breaking changes for v7.0

### LINKED ISSUE #925: Change the AppName from "CP Editor" to "cpeditor"
. The setting files will be placed in "xxx/xxx/xxx/cpeditor" instead of "xxx/xxx/xxx/CP Editor", and we'll move the setting files. This will also cause some other changes, for example, the class in xprop will change.

ouuan:
## Description

#769 except for YAML settings. Also, drop legacy setting migration from v6.7 and earlier.

takes over #934, closes #925 and #926.

## How Has This Been Tested?

On Arch Linux.

## Checklist


ouuan:
These changes were planned to be merged into v7.0 because new releases before it like v6.12 were expected. Now they can just be merged into master.

src/Settings/SettingsUpdater.cpp
coder3101, line 1: What is the reason for this change?

ouuan: They are pretty old. I think there is no need to always keep them.

coder3101 (approved):

Issue #1322, opened 2024-02-18
Why not use KDE dependencies 


yurivict:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this feature provided in the master branch?

- [X] I have checked the changelog

### The problem you are facing

Why are these git modules used:
```
[submodule "third_party/extra-cmake-modules"]
	path = third_party/extra-cmake-modules
	url = https://github.com/KDE/extra-cmake-modules.git
[submodule "third_party/syntax-highlighting"]
	path = third_party/syntax-highlighting
	url = https://github.com/KDE/syntax-highlighting.git
```
?

Why not use packages for these which are part of KDE and are available everywhere?

### Describe the feature you'd like

Use external packages when possible.

### Are you willing to contribute to this?

- [ ] I'm ready to contribute to this (and will open a PR soon)
- [ ] I'd like to have a try (with the help of the maintainers)
- [X] No, thanks

ouuan:
I did suggest that don't include them as submodules, but it eventually got merged without this suggestion being resolved. Now on Arch Linux (AUR) the system package is used, but not for Windows/AppImage/macOS.

ouuan:
Yes, we can fix this now, but I don't think it causes too much trouble for packaging. Maybe some automation script will clone all submodules, causing some waste, but otherwise it should be just fine.

ouuan:
But I'm not sure whether we should use the system library in AppImage. The version on Ubuntu 20.04 is very old 🤔 And I'm not familiar with packaging on Deb/macOS, so I think modifications are left to other contributors.

But now I can remove them from the submodules and just clone from upstream in CI.

yurivict:
They are just unnecessary.

ouuan:
> They are just unnecessary.

Yes, I said that I agree with that. 

> but I don't think it causes too much trouble for packaging

I mean a new release is not needed to enable packaging with the system library.

coder3101:
For CI, let's not pull upstream. If we are going system package dependency then let's use whatever stable releases the current OS has to offer.

For macOS, there is a `homebrew/tap` for linux package managers have this, I am not sure for windows, there's an installer for installing KDE frameworks.

ouuan:
There are several issues here:

-   Putting ECM and KSH in submodules is misleading. It suggests that these versions should always be used instead of the system library. This can be resolved by removing them from the submodules.
[...]

PR #1323, opened 2024-02-18
chore: remove ECM and KSH submodules

ouuan:
## Description

Remove ECM and KSH submodules and clone them in CI.

#1322


coder3101:
We should not build the KSH or ECM in CI rather use package managers to install them. We are going with upgrading to qt6 and `KF5*` packages such as `syntax-higlighting` needs qt5.

- For macOS we can use [this official homebrew  tap](https://invent.kde.org/packaging/homebrew-kde)
- For Ubuntu this package should work `libkf5syntaxhighlighting-dev`
- For Windows, we can this [craft](https://community.kde.org/Get_Involved/development/Windows) but I haven't tried this nor do I have a windows machine.

coder3101:
macOS builds are failing because of architecture mismatch. GH has rolled out `arm64` with `macos-latest`

coder3101:
As first step it's good to just get rid of the submodules.

coder3101 (approved):

PR #1490, opened 2026-05-17
fix(snap): use DESTDIR to stage KF6 SyntaxHighlighting into snap

coder3101:
## Description
Fixes `libKF6SyntaxHighlighting.so.6: cannot open shared object file` error when running the core26 snap.

Without `DESTDIR=$CRAFT_PART_INSTALL`, `cmake --install` installs ECM and syntax-highlighting to the system `/usr` inside the build instance instead of the craft part install directory. This means the libraries are never staged into the final snap.

Fixes runtime error after snap core26 migration (#1489)

## Checklist
- [ ] If the key of a setting is changed, the `old` attribute is updated or it is resolved in SettingsUpdater.
- [ ] If there are changes of the text displayed in the UI, they are wrapped in `tr()` or `QCoreApplication::translate()`.
- [ ] If needed, I have opened a pull request or an issue to update the [documentation](https://github.com/cpeditor/cpeditor.github.io).

### DIFF
--- a/.github/workflows/push_snap.yml
+++ b/.github/workflows/push_snap.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - fix/snap-destdir-staging
     paths-ignore:
       - ".github/**"
       - "!.github/workflows/push_snap.yml"
--- a/.github/workflows/push_snap.yml
+++ b/.github/workflows/push_snap.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


.github/workflows/push_snap.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - fix/snap-destdir-staging
```
coder3101:
```suggestion
```

asharkhan3101 (approved):

